PR #6, opened 2020-05-12
docs(spec): Custom Security Resources Specification

danielpacak:


danielpacak:
Let's do that when the API is stable.

PR #54, opened 2020-06-11
feat: Add docker integration

sylus:
Hi there ^_^ 

Was wondering if could add docker integration for starboard so I can run off official image and run it as a cronjob?

Thanks so much love the tool!

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,34 @@
+FROM golang:1.14.0 AS build
+WORKDIR /go/src/github.com/aquasecurity/starboard/
+COPY go.mod go.sum ./
+COPY kube/ kube/
+COPY pkg/ pkg/
+COPY cmd/ cmd/
+ARG STARBOARD_VERSION
+RUN GO111MODULE=on CGO_ENABLED=0 go build -o ./bin/starboard ./cmd/starboard/main.go
+
+FROM alpine:3.11 AS run
+WORKDIR /opt/starboard/
+# add GNU ps for -C, -o cmd, and --no-headers support
+RUN apk --no-cache add procps
+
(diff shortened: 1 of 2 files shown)


danielpacak:
Hi @sylus ! Thank you for opening this PR. I'm also glad that you liked the tool!

I think that having an official Docker image for Starboard might be useful. However I'd consider integrating it with GoReleaser, that we use to publish executable binaries to GitHub releases (see https://goreleaser.com/customization/docker/) By doing that we could publish up to date Docker images to DockerHub whenever a new release tag is created. Similar to what we do for other projects, e.g. https://github.com/aquasecurity/harbor-scanner-trivy/releases

Having said that, in your use case, do you intend to build an image in your environment or would you rather pull it from a registry such as DockerHub and run Starboard commands?

danielpacak (changes requested):
I think it's a good idea to publish Starboard as Docker image. However we should integrate it with our release pipeline which relies on GoReleaser (see https://goreleaser.com/customization/docker/)

@sylus Are you available to work on this one and configure Docker build similarly to what we do in https://github.com/aquasecurity/harbor-scanner-trivy/blob/master/.goreleaser.yml#L29

danielpacak:
@sylus We've opened another PR to resolve this as we need it for some other integration, see #152

danielpacak:
Starting with release v0.4.0 it's possible to run Starboard as Docker container:

```
$ docker run docker.io/aquasec/starboard:0.4.0 version
Starboard Version: {Version:0.4.0 Commit:dd8e49701c1817ea174061c8731fe5bdbfb73d93 Date:2020-09-21T09:36:59Z}
```

Issue #58, opened 2020-06-16
Automate krew plugin updates


danielpacak:
**Describe the problem/challenge you have**

Each time there's a new release of Starboard CLI, we have to bump up versions and checksums in the krew manifest and open a PR to the krew-index repository. This is repetitive and mundane task for us and krew-index maintainers.

**Describe the solution you'd like**

Integrated [krew-release-bot](https://krew.sigs.k8s.io/docs/developer-guide/release/automating-updates/) with the GitHub Actions release workflow.

**Anything else you would like to add:**

IIRC We have to submit the initial manifest to krew-index manually, i.e. we can work on that only after https://github.com/kubernetes-sigs/krew-index/pull/647/files is merged.

**Environment:**

- Starboard version (use `starboard version`):
- Kubernetes version (use `kubectl version`):
- OS (macOS 10.15, Windows 10, Ubuntu 19.10 etc):

krol3:
I'd like to help with this issue.

evalsocket:
Hey, @krol3 It looks pretty simple and easy. I also have some idea about krew. Can this assign to me if you are not working on it

krol3:
@evalsocket  I submitted the PR, could you take a look? If I'm missing something.

PR #72, opened 2020-06-24
chore: script to use krew-release-bot

krol3:
About Issue https://github.com/aquasecurity/starboard/issues/58 was added the file  .krew.yaml to integrate with krew-release-bot 

#### Testing 

```
docker run --mount type=bind,source=$PWD/.krew.yaml,target=/tmp/krew.yaml rajatjindal/krew-release-bot:v0.0.38 \
  krew-release-bot template --tag v0.2.2 --template-file /tmp/krew.yaml
```

image


danielpacak:
@krol3 @evalsocket I think we missed one step to actually send an event to bot whenever a new release is performed. See for example https://github.com/ahmetb/kubectl-tree/blob/master/.github/workflows/release.yml#L23

krol3:
The workflow syntax for Github actions works in .github, I didn't notice :( . I updated the PR

danielpacak:
I pushed one more commit to make the krew manifest template as close to the existing one as possible https://github.com/kubernetes-sigs/krew-index/blob/master/plugins/starboard.yaml

Otherwise it would not get merged automatically.

PR #76, opened 2020-07-02
docs(proposal): Introduce a custom resource to represent a (trusted) registry

danielpacak:

### DIFF
--- a/proposals/trusted_registry_crd.md
+++ b/proposals/trusted_registry_crd.md
@@ -108,8 +108,18 @@ report:
     noneCount: 0
     unknownCount: 3
   vulnerabilities:
-  - id: CVE-2020-1879
-  - id: CVE-2018-2342
+    - vulnerabilityID: CVE-2020-3810
+      resource: apt
+      severity: MEDIUM
+      installedVersion: 1.8.2
+      fixedVersion: 1.8.2.1
+      description: Missing input validation in the ar/tar implementations of APT before
+        version 2.1.2 could result in denial of service when processing specially crafted
+        deb files.
+      links:
+        - https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2020-3810
+        - https://github.com/Debian/apt/issues/111
+    - vulnerabilityID: CVE-2018-2342
 ```
 
 ## Starboard Security Operator supervising Registry entities
--- a/proposals/trusted_registry_crd.md
+++ b/proposals/trusted_registry_crd.md
@@ -68,7 +68,7 @@ status:
     lastScanTime: null
 ```
 
-The `Registry` entity has the `spec` property, which allows specifying trusted repositories. The `status` filed might
+The `Registry` entity has the `spec` property, which allows specifying trusted repositories. The `status` field might
 contain scan status for individual repositories. In the example above, you can see that the `library/nginx` repository
 has already been scanned, whereas the `library/alpine` repository has not.
 
--- a/proposals/trusted_registry_crd.md
+++ b/proposals/trusted_registry_crd.md
@@ -16,7 +16,7 @@ Common requirements for running secure Kubernetes workloads are:
 - Pull container images only from trusted registries and/or repositories. For example, one might allow pulling
   images from private Harbor repository accessible at https://core.harbor.domain/library/nginx:1.16, but deny pull
   requests to the core.harbor.domain/library/redis:5 or https://docker.io/library/nginx:1.16 repositories.
-- Statically scan container images for critical, high or medium vulnerabilities. In general, these rules are configurable.
+- Statically scan container images, and implement rules to deny workloads that include critical, high or medium vulnerabilities. In general, these rules are configurable (for example, what level of severity will cause an image to be denied).
 - Use a Kubernetes admission controller to deny workloads with non-compliant container images. For example, one may
   allow `kubectl run nginx-pod --image=core.harbor.domain/library/nginx:1.16`, but deny
   `kubectl run redis-pod --image core.harbor.domain/library/redis:5`.
--- a/proposals/trusted_registry_crd.md
+++ b/proposals/trusted_registry_crd.md
@@ -1,0 +1,128 @@
+# Introduce a custom resource to represent a (trusted) registry
+
+## TOC
+
+- [Abstract](#abstract)
+- [Problem statement](#problem-statement)
+- [Solution](#solutions)
+- [Use cases](#use-cases)
+  - [Starboard Security Operator supervising trusted registries](#starboard-security-operator-supervising-trusted-registries)
+  - [Repository webhooks integrated with Starboard](#repository-webhooks-integrated-with-starboard)
+
+## Abstract
+
+Common requirements for running secure Kubernetes workloads are:
+
+- Pull container images only from trusted registries and/or repositories. For example, one might allow pulling
+  images from private Harbor repository accessible at https://core.harbor.domain/library/nginx:1.16, but deny pull
+  requests to the core.harbor.domain/library/redis:5 or https://docker.io/library/nginx:1.16 repositories.
+- Statically scan container images for critical, high or medium vulnerabilities. In general, these rules are configurable.
+- Use a Kubernetes admission controller to deny workloads with non-compliant container images. For example, one may
+  allow `kubectl run nginx-pod --image=core.harbor.domain/library/nginx:1.16`, but deny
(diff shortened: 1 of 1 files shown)


proposals/trusted_registry_crd.md
```diff
@@ -0,0 +1,128 @@
...
+  - [Starboard Security Operator supervising trusted registries](#starboard-security-operator-supervising-trusted-registries)
+  - [Repository webhooks integrated with Starboard](#repository-webhooks-integrated-with-starboard)
+
+## Abstract
+
+Common requirements for running secure Kubernetes workloads are:
+
+- Pull container images only from trusted registries and/or repositories. For example, one might allow pulling
+  images from private Harbor repository accessible at https://core.harbor.domain/library/nginx:1.16, but deny pull
+  requests to the core.harbor.domain/library/redis:5 or https://docker.io/library/nginx:1.16 repositories.
```
lizrice, line 18: I might be leaping to wrong assumptions here but I think the tag is irrelevant? Suggested changes for clarification (but maybe I have the wrong end of the stick?)

```suggestion
  images from private Harbor repository accessible at https://core.harbor.domain/library/nginx, but deny pull
  requests to https://core.harbor.domain/library/redis (unapproved repo `redis`) or https://docker.io/library/nginx (unapproved registry `docker.io`).
```

proposals/trusted_registry_crd.md
```diff
@@ -0,0 +1,128 @@
...
+- Statically scan container images for critical, high or medium vulnerabilities. In general, these rules are configurable.
```
lizrice, line 19:
```suggestion
- Statically scan container images, and implement rules to deny workloads that include critical, high or medium vulnerabilities. In general, these rules are configurable (for example, what level of severity will cause an image to be denied).
```

proposals/trusted_registry_crd.md
```diff
@@ -0,0 +1,128 @@
...
+  report associated with a given container image digest, for example `registryvulnerabilities.aquasecurity.github.io`.
+- Container images are not first-class citizens in Kubernetes-native world. There’s no built-in resource that represents
+  a container image. Also, there’s no built-in representation of a (trusted) registry.
+- An admission controller has certain performance requirements. Usually we cannot trigger a scanner when the admission
+  controller is invoked for a given Kubernetes API request. Most likely it will take way too long to perform such
+  ad hoc scans.
+
+## Solutions
+
+### Define registries.aquasecurity.github.io resource
```
lizrice, line 39: Nit - should this be `registry` in the singular (like `pod` or `node`)? I think this differs from vulnerabilities because we have many vulns in one instance

proposals/trusted_registry_crd.md
```diff
@@ -0,0 +1,128 @@
...
+spec:
+  name: My Harbor Registry
+  description: |
+    That's the only registry authorized to run production workloads.
+  server: core.harbor.domain
+  imagePullSecrets:
+  - name: my-harbor-registry-credentials
+  insecure: false
+  trustedRepositories:
+  - name: library/nginx:1.16
```
lizrice, line 58: Do you think we need to restrict down to the level of approved / not-approved tags? I was imagining more at the repo level. Or even just to the registry level. 

Registries - only allow images from registries that have a CRD instance in place 
Repositories - do they have to have a trustedRepositories entry? Or maybe the default is to allow everything from all repos in a trusted registry? (I am not totally convinced we need the granularity of repository-level control for the first version...) 

Edit: on further reading I think you are suggesting using this as the list of repos that the Starboard operator knows to scan. I wonder if this is conflating two different things. We could have a CRD for trusted registries, and a Starboard config map for a list of repos to be scanned? I am not sure if this is better, just raising it as an option. 

Having a CRD for trusted registries would allow for a super-simple admission controller that denies images that don't come from a trusted registry.

proposals/trusted_registry_crd.md
```diff
@@ -0,0 +1,128 @@
...
+  trustedRepositoryStatuses:
+  - name: library/nginx
+    scanned: true
+    lastScanTime: 2020-01-01T00:00:00Z
+  - name: library/alpine
+    scanned: false
+    lastScanTime: null
+```
+
+The `Registry` entity has the `spec` property, which allows specifying trusted repositories. The `status` filed might
```
lizrice, line 71:
```suggestion
The `Registry` entity has the `spec` property, which allows specifying trusted repositories. The `status` field might
```

proposals/trusted_registry_crd.md
```diff
@@ -0,0 +1,128 @@
...
+contain scan status for individual repositories. In the example above, you can see that the `library/nginx` repository
+has already been scanned, whereas the `library/alpine` repository has not.
```
lizrice, line 73: What if the repo is part-scanned? Some image versions might be scanned, others might not...

proposals/trusted_registry_crd.md
```diff
@@ -0,0 +1,128 @@
...
+      mimeType: "application/vnd.docker.distribution.manifest.v2+json"
+  summary:
+    criticalCount: 0
+    highCount: 0
+    lowCount: 92
+    mediumCount: 27
+    noneCount: 0
+    unknownCount: 3
+  vulnerabilities:
+  - id: CVE-2020-1879
```
knqyf263, line 111: Did you just omit details such as a package name and CVSS? They are necessary so that a complicated policy can be applied in Admission Controller.

proposals/trusted_registry_crd.md
```diff
@@ -0,0 +1,128 @@
...
+## Solutions
+
+### Define registries.aquasecurity.github.io resource
+
+First off, introduce a custom resource to represent a (trusted) container registry. For example, we can represent
+a private Harbor registry with the following Kubernetes object:
+
+```yaml
+apiVersion: aquasecurity.github.io/v1alpha1
+kind: Registry
```
knqyf263, line 46: Is this resource required to create Registry Vulnerability? I think it is unnecessary for some users to specify trusted registries. They might just want to audit vulnerabilities in the image they're trying to deploy.

danielpacak: I agree. I think we should be able to create vulnerability reports for the specified image digest and store it at the cluster level without requiring the registry resource.

proposals/trusted_registry_crd.md
danielpacak: Usually the names that you get from `k api-resources --api-group ""` are plural. Also fully qualified resource names must be plural in CRD descriptors:

```
k explain crd.spec.names.plural
KIND:     CustomResourceDefinition
VERSION:  apiextensions.k8s.io/v1

FIELD:    plural <string>

DESCRIPTION:
     plural is the plural name of the resource to serve. The custom resources
     are served under `/apis/<group>/<version>/.../<plural>`. Must match the
     name of the CustomResourceDefinition (in the form
     `<names.plural>.<group>`). Must be all lowercase
```

proposals/trusted_registry_crd.md
danielpacak: We'll have all the details as we have for `vulnerabilities.aquasecurity.github.io`. For example, https://github.com/aquasecurity/starboard/blob/master/kube/examples/vulnerabilities.yaml#L30 I've update the sample snippet.

proposals/trusted_registry_crd.md
danielpacak: - The first use case for such CRD that came to my mind was an admission controller that required a container image to be already scanned. For example, the admission controller could reject a deployment if:
  - the container image hasn't been scanned
  - the container image has already been scanned but has a critical vulnerability
- In other words, my assumption was that we do reject deployments by default, unless they're scanned
- The other use case for the registries CRD is to specify the desired state of which container images we want to be scanned. Whereas the actual state is represented by the corresponding vulnerability reports. So the desired state could be reconciled by the Starboard Security Operator if the integration via webhooks was not possible

danielpacak: And I agree, we should think more about the granularity of such allow list, whether it's registry, repository, or tag/digest level.

proposals/trusted_registry_crd.md
danielpacak: I think that eventually an admission controller might be checking existence of the registryvulnerabilities reports for the given deployment request. The status at the registry level would be just a summary / current status of the reconciliation loop.

danielpacak:
We put more thoughts into that. Storing all possible vulnerability reports seems more like a function for a registry or for a security solution like Aqua - let's keep Starboard for "live" security reports about what's happening in the cluster.

PR #78, opened 2020-07-05
feat(crd): add ContainerVulnerabilities

knqyf263:
# Overview
Ref. https://github.com/aquasecurity/starboard/pull/76

# ToDo
- [x] create CRD yaml
- [x] code-generate
- [x] initialize ContainerVulnerabilities in `init` subommand
- [x] support Artifact and Registry
- [x] support `image/` in `find vulnerabilities` subcommand
- [ ] support `imagePullSecrets`
- [ ] which namespace does the job run in?

# Note
I've just prepared PoC so that we can discuss how to implement it easily.


pkg/find/vulnerabilities/crd/writer.go
```diff
@@ -49,3 +55,17 @@ func (s *writer) createVulnerability(ctx context.Context, workload kube.Object,
 
 	return err
 }
+
+func (s *writer) createContainerVulnerability(ctx context.Context, workload kube.Object, report starboard.VulnerabilityReport) (err error) {
+	_, err = s.client.AquasecurityV1alpha1().ContainerVulnerabilities().Create(ctx, &starboard.ContainerVulnerability{
+		ObjectMeta: meta.ObjectMeta{
+			Name: fmt.Sprintf(uuid.New().String()),
+			Labels: map[string]string{
+				kube.LabelResourceKind: string(workload.Kind),
```
knqyf263, line 64: Should we add more labels?

pkg/find/vulnerabilities/trivy/converter.go
```diff
@@ -7,14 +7,15 @@ import (
...
+	"github.com/google/go-containerregistry/pkg/name"
 )
 
 // Converter is the interface that wraps the Convert method.
 //
 // Convert converts the vulnerabilities model used by Trivy
 // to a generic model defined by the Custom Security Resource Specification.
 type Converter interface {
-	Convert(reader io.Reader) (starboard.VulnerabilityReport, error)
+	Convert(imageRef string, reader io.Reader) (starboard.VulnerabilityReport, error)
```
knqyf263, line 18: Can I change the interface?

pkg/find/vulnerabilities/trivy/converter.go
```diff
@@ -37,6 +38,14 @@ func (c *converter) Convert(reader io.Reader) (report starboard.VulnerabilityRep
 		return
 	}
 	report = c.convert(scanReports)
+
+	ref, err := name.ParseReference(imageRef)
```
knqyf263, line 42: If you don't want to add more dependency, let me know, please. `moby/moby` also provides such a function, but it is heavier than `google/go-containerregistry`.

pkg/find/vulnerabilities/trivy/scanner.go
```diff
@@ -226,7 +231,9 @@ func (s *Scanner) GetVulnerabilityReportsByScanJob(ctx context.Context, job *bat
 		if err != nil {
 			return
 		}
-		reports[c.Name], err = s.converter.Convert(logReader)
+
+		imageRef := c.Args[len(c.Args)-1]
```
knqyf263, line 235: It is kind of workaround because `job.Spec.Template.Spec.Containers` doesn't have the field representing the scanned image name.

pkg/kube/object.go
```diff
@@ -23,6 +23,9 @@ const (
 
 	KindNode Kind = "Node"
 
+	// pseudo Kind
+	KindImage Kind = "Image"
```
knqyf263, line 27: If you don't prefer defining `KindImage` since it is not a Kubernetes resource, tell me any better way. I'm definitely sure that you are more familiar than me.

knqyf263:
We decided not to support this feature.

PR #81, opened 2020-07-08
fix(trivy): Set Registry and Artifact properties in the VulnerabilityReport

danielpacak:


pkg/find/vulnerabilities/trivy/converter.go
```diff
@@ -79,15 +79,22 @@ func (c *converter) convert(reports []ScanReport) starboard.VulnerabilityReport
 		}
 	}
 
+	registry, artifact, err := c.parseImageRef(imageRef)
+	if err != nil {
+		return starboard.VulnerabilityReport{}, nil
```
knqyf263, line 84: Do we return `nil`? Even if so, we might need kind of logging.

danielpacak: Good catch. I meant to return the error. It's fixed now with a test case to assert that the error is indeed returned

Issue #45, opened 2020-06-03
Run kube-bench on all cluster nodes


danielpacak:
Currently the kube-bench job is only scheduled on a single worker node. However, we should run it for all nodes. What's more we should tell kube-bench to run different checks on master and worker nodes respectively <sup>[1]</sup> .

[1]: https://github.com/aquasecurity/kube-bench#running-in-a-kubernetes-cluster

PR #91, opened 2020-07-13
chore - kube-bench in all nodes

krol3:
Issue https://github.com/aquasecurity/starboard/issues/45 run kube-bench in all nodes


pkg/cmd/kube_bench.go
```diff
@@ -30,15 +37,25 @@ func NewKubeBenchCmd(cf *genericclioptions.ConfigFlags) *cobra.Command {
...
 			if err != nil {
+				err = fmt.Errorf("list nodes: %w", err)
 				return
 			}
-			err = crd.NewWriter(ext.NewSystemClock(), starboardClientset).Write(ctx, report, node)
+			for _, nodeItem := range nodeList.Items {
+				klog.V(3).Infof("Node name: %s/%s", nodeItem.Name, nodeItem.Labels[masterNodeLabel])
+				report, node, err := kubebench.NewScanner(opts, kubernetesClientset).Scan(ctx, nodeItem.Name)
+				if err != nil {
+					break
```
danielpacak, line 54: I think it's better to klog.Warnf to see why scanning has failed

pkg/kubebench/scanner.go
```diff
@@ -26,6 +26,7 @@ import (
 const (
 	kubeBenchVersion       = "0.3.0"
 	kubeBenchContainerName = "kube-bench"
+	masterNodeLabel        = "node-role.kubernetes.io/master"
```
danielpacak, line 29: Is this const used somewhere? I think we might want to use kube-bench node type detection feature

pkg/cmd/kube_bench.go
```diff
@@ -30,15 +37,25 @@ func NewKubeBenchCmd(cf *genericclioptions.ConfigFlags) *cobra.Command {
...
+			// List Nodes
+			nodeList, err := kubernetesClientset.CoreV1().Nodes().List(ctx, meta.ListOptions{})
 			if err != nil {
+				err = fmt.Errorf("list nodes: %w", err)
 				return
 			}
-			err = crd.NewWriter(ext.NewSystemClock(), starboardClientset).Write(ctx, report, node)
+			for _, nodeItem := range nodeList.Items {
+				klog.V(3).Infof("Node name: %s/%s", nodeItem.Name, nodeItem.Labels[masterNodeLabel])
+				report, node, err := kubebench.NewScanner(opts, kubernetesClientset).Scan(ctx, nodeItem.Name)
```
danielpacak, line 52: The Scan() method is blocking and it takes some time to run the benchmarks. Instead of running the tests sequentially, I'm wondering if we can use goroutines and wait group to run tests in parallel, i.e. run a goroutine per cluster node.

pkg/kubebench/scanner.go
krol3: Yes! We could use to identify parameters in kube-bench

PR #98, opened 2020-07-27
test(kube-bench): Add integration test for kube-bench command

danielpacak:
This PR:
- Adds scenario for integration test kube-bench command
- Attempts to test kube-hunter in a similar way to kube-bench, but I gave up for now. The test is failing on GitHub actions runner for some reason.
- Fixes bug in kube-bench runner, where we called wg.Done() too early. This caused an issue with CISKubeBenchReports which where not sent to Kebernetes API

/cc @krol3 


pkg/kubebench/scanner.go
```diff
@@ -102,18 +102,22 @@ func (s *Scanner) Scan(ctx context.Context, nodeName string, target string, wg *
...
 	return &batch.Job{
 		ObjectMeta: meta.ObjectMeta{
 			Name:      uuid.New().String(),
 			Namespace: kube.NamespaceStarboard,
-			Labels: map[string]string{
-				"app": "kube-bench",
+			Labels: labels.Set{
+				"app.kubernetes.io/name": "starboard-cli",
+				kube.LabelResourceKind:   string(kube.KindNode),
+				kube.LabelResourceName:   node.Name,
```
lizrice, line 120: I wonder if in addition we should add a label to indicate that it's a kube-bench job? something like `"starboard-app": "kube-bench"` maybe?  (also for pod-spec and for other kinds of tool)

itest/integration_test.go
```diff
@@ -0,0 +1,134 @@
...
+		command := exec.Command(pathToStarboardCLI, []string{"cleanup", "-v", "3"}...)
+		session, err := Start(command, GinkgoWriter, GinkgoWriter)
+		Expect(err).ToNot(HaveOccurred())
+		Eventually(session).Should(Exit(0))
+
+		// TODO We have to wait for the termination of the starboard namespace. Otherwise the init command
+		// TODO run by the BeforeEach callback fails when it attempts to create Kubernetes objects in the
+		// TODO starboard namespace that is being terminated.
+		//
+		// TODO Maybe the cleanup command should block and wait unit the namespace is terminated?
```
lizrice, line 127: I think this might be a good idea for users too. I ran into this recently - cleanup followed quickly by init failed. It was pretty obvious why, but waiting for the namespace to be gone would prevent it from being a problem at all.

pkg/kubebench/scanner.go
danielpacak: Yes. That's a very good point. I'll take it as a separate issue to review the labels to identify scan job and the underlying tools.
I think it's less relevant for Starboard CLI, but the Operator has no clue which scanner we're running in the starboard namespace.

itest/integration_test.go
danielpacak: Created a task to follow up https://github.com/aquasecurity/starboard/issues/99

Issue #101, opened 2020-07-30
Create or update CISKubeBenchReport resources


danielpacak:
**Describe the problem/challenge you have**

In the initial version of Starboard we wanted to store historical benchmark reports. However, now we think that we should store only the latest report as a CR instance.

**Describe the solution you'd like**

Simplify the logic and implement create or update logic for CISKubeBenchReport writer.

**Anything else you would like to add:**

N/A

**Environment:**

- Starboard version (use `starboard version`): 0.2.6
- Kubernetes version (use `kubectl version`): Any
- OS (macOS 10.15, Windows 10, Ubuntu 19.10 etc): Any

PR #116, opened 2020-08-12
feat: Create or update CISKubeBenchReports

danielpacak:
In the initial version of Starbaord we wanted to store historical
benchmark reports. However, we now think that it's more appropriate
to store only the latest report.

Resolves: #101

### DIFF
--- a/pkg/kubebench/crd/writer.go
+++ b/pkg/kubebench/crd/writer.go
@@ -2,6 +2,7 @@ package crd
 
 import (
 	"context"
+
 	"github.com/aquasecurity/starboard/pkg/generated/clientset/versioned/typed/aquasecurity/v1alpha1"
 	"k8s.io/klog"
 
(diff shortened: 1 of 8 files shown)


pkg/kubebench/crd/writer.go
```diff
@@ -76,69 +49,24 @@ func (w *writer) Write(ctx context.Context, report starboard.CISKubeBenchOutput,
...
-func (w *writer) removeHistoryLatestLabel(ctx context.Context, reports []starboard.CISKubeBenchReport) (err error) {
-	for _, report := range reports {
-		if value, ok := report.Labels[kube.LabelHistoryLatest]; !ok || value != "true" {
-			continue
-		}
-		clone := report.DeepCopy()
-		delete(clone.Labels, kube.LabelHistoryLatest)
-		klog.V(3).Infof("Removing %s label from %s report", kube.LabelHistoryLatest, clone.Name)
-		_, err = w.clientset.AquasecurityV1alpha1().CISKubeBenchReports().Update(ctx, clone, meta.UpdateOptions{})
+	if errors.IsAlreadyExists(err) {
```
knqyf263, line 52: What if we always remove the existing resource before trying to create a new report or call `w.reports.Get` at first? I'm not sure I understand properly, but this report is generated for each node, right? In that case, creating a new report always fails after the second time. The current implementation calls 3 APIs if the report already exists, Create → Get → Update. I think we can call only 2  APIs, Delete → Create, or (Get → Create || Get → Update). The bad point of Delete → Create is that the report cannot be accessed only for a moment.

It may not make sense to reduce the number of API calls, but I think fewer is better than many. It also reduces source code.

danielpacak: Yes. That's a very good point @knqyf263 I was thinking about that while implementing the change, but you convinced me that we might want to Get the report first because most of the time we'll be updating an existing report.

We don't want to Delete the report to handle the case when Create fails, which is quite likely in the distributed system

knqyf263: >because most of the time we'll be updating an existing report.

Yes, that is exactly what I wanted to tell you. You understand what I meant better than me 😆 

>We don't want to Delete the report to handle the case when Create fails, which is quite likely in the distributed system

Right. The system which is able to accept the case sometimes goes with the way for simplicity. In this case, it is probably better to call `Get` first.

danielpacak: Okey. I've changed the code to Get and then either Create (if error was not found) or Update (if the report was there)

PR #118, opened 2020-08-13
Feature: Single workload scanning with polaris

### LINKED ISSUE #29: Run Polaris for the specified workload
**Describe the problem/challenge you have**

Currently, we're running [Polaris] audit subcommand to scans all workloads in the cluster.

```
$ starboard polaris
$ starboard get configaudit deploy/nginx -n dev
```

**Describe the solution you'd like**

However, we should be able to scan a single workload like we do for vulnerabilities:

```
$ starboard polaris deploy/nginx -n dev
$ starboard polaris sts/my-app -n staging
```

- This is mainly required for multi tenant environments and workloads protected by RBAC permissions. Beyond that, scanning all workloads requires running Polaris as Kubernetes Job with ServiceAccount that's not least privileged.

[polaris]: https://github.com/FairwindsOps/polaris

danielsagi:
Due to the new release of the single workload scanning method of polaris (in version 1.2),
This PR integrates this feature.
From now on, instead of scanning the entire cluster for config audits, you just run:
```bash
$ kubectl starboard polaris deploy/my-deployment
```

### Internal change
This introduced a change in the way we map aliases of kinds to full kinds QN
Up until now we used a switch case for resolving kind names. this method is not robust enough and can fail in some situations.
The correct way to do that, (and what this PR is introducing) is to use [RESTMapper](https://godoc.org/k8s.io/client-go/restmapper)

Other than the general correctness of using a Mapper instead of a switch case,
We needed a way to figure out the api version and group name for kinds/resources when we pass the resource argument to polaris.

#### Tests change
Because we now use a `RESTMapper` rather than a static function. the test for resolving kinds has changed to be an integration test.

closes #29


pkg/kube/object.go
```diff
@@ -25,13 +25,13 @@ const (
 	KindNode Kind = "Node"
 
 	KindPod                   Kind = "Pod"
-	KindReplicaSet            Kind = "ReplicaSet"
+	KindReplicaSet            Kind = "ReplicaSet.apps"
```
danielpacak, line 28: As we discussed we cannot mix Kubernetes Kinds with fully qualified resource names.

danielpacak (approved):

Issue #69, opened 2020-06-22
Notarize starboard through Apple Developer channel


danielpacak:
**Describe the problem/challenge you have**

On macOS running starboard CLI requires adding the binary to security exceptions.

**Describe the solution you'd like**

We should notarize our binaries as trusted provider and integrate such process with GoReleaser

**Anything else you would like to add:**

N/A

**Environment:**

- Starboard version (use `starboard version`): 0.2.1
- Kubernetes version (use `kubectl version`): Any
- OS (macOS 10.15, Windows 10, Ubuntu 19.10 etc): macOS

lizrice:
On consideration it doesn't seem worth doing this at present, it doesn't appear that other Kubernetes tools go through this process

PR #119, opened 2020-08-13
notarization: Add initial stuff for gon integration with goreleaser

simar7:
This is some initial work to make gon work with goreleaser and release notarized release artifacts for consumption.

Addresses: https://github.com/aquasecurity/starboard/issues/69

Signed-off-by: Simarpreet Singh <simar@linux.com>

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -6,18 +6,38 @@ on:
 jobs:
   release:
     name: Release
-    runs-on: ubuntu-18.04
+    runs-on: macos-latest
     steps:
       - name: Setup Go
         uses: actions/setup-go@v1
         with:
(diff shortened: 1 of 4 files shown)


simar7:
I've added the appropriate secrets to this repo for this to work. In theory it should work as it does locally but I haven't had the chance to test this with GitHub Actions just yet.

simar7:
There are some limitation of running gon in a CI environment where user input is not available: https://github.com/mitchellh/gon#prompts

For this reason I think we might have to do something like this: https://github.com/paginagmbh/EPUB-Checker/blob/master/src/build/mac-release.sh to manually do some steps to get gon going along. I haven't tested this but if the current setup doesn't work this where I would look next.

Issue #26, opened 2020-05-22
Set ownerReference for vulnerabilities resource


danielpacak:
By design vulnerabilities are associated with native Kubernetes workloads such as Deployments or Stateful Sets via labels and label selectors. To leverage Kubernetes Garbage Collection we can set the ownerReference for each vulnerabilities resource when we create it.

krol3:
I'd like to help it

danielpacak:
@krol3 To avoid conflicts, let's wait for https://github.com/aquasecurity/starboard/pull/95 to be merged

PR #129, opened 2020-08-24
feat: Set ownerReference for vulnerabilities resource

danielpacak:
Resolves: #26

### DIFF
--- a/pkg/kube/object.go
+++ b/pkg/kube/object.go
@@ -4,6 +4,10 @@ import (
 	"encoding/json"
 	"fmt"
 
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+
 	"k8s.io/apimachinery/pkg/labels"
 )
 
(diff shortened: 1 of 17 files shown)


pkg/kube/object.go
```diff
@@ -84,3 +88,113 @@ func (ci ContainerImages) AsJSON() (string, error) {
...
+	return nil
+}
+
+func upsertOwnerRef(ref metav1.OwnerReference, object metav1.Object) {
+	owners := object.GetOwnerReferences()
+	idx := indexOwnerRef(owners, ref)
+	if idx == -1 {
+		owners = append(owners, ref)
+	} else {
+		owners[idx] = ref
```
knqyf263, line 172: Just for my understanding:
In this case, what field will be updated? Only `UID` considering the unit test? I'm curious when it happens. It looks like `Group`, `Kind`, and `Name` are always the same.

knqyf263 (approved):

pkg/kube/object.go
danielpacak: In this case all properties of the given `metav1.OwnerReference` are updated including GVK even though it's very unlikely they change in real Starboard scenario. However this utility method is generic to cater for cases where we have some other controller / Kubernetes app that also relies on the `ownerReferences` property. In this case the GVK might be different.

knqyf263: If `indexOwnerRef` returns `idx` other than `-1`, it means GVK is the same, right? So, I wanted to know why `owners[idx] = ref` is necessary. If all properties are the same as existing `ownerReference`, we don't need to update it. But `UID` might be different when a user deletes the resource and recreates it with the same GVK and name.

danielpacak: Yup. You're right. It might only happen if the UID has changed.

knqyf263: I see. Thank you for the clarification!

PR #135, opened 2020-08-27
Update Makefile to add a test target

krisctl:


krisctl:
@danielpacak PTAL, thanks!

Makefile
```diff
@@ -24,6 +24,10 @@ compile-templates: get-qtc
 $(BINARY): $(SOURCES)
 	CGO_ENABLED=0 go build -o ./bin/$(BINARY) ./cmd/starboard/main.go
 
+.PHONY: test
+## test will run both unit tests and integration tests
+test: unit-tests integration-tests
```
danielpacak, line 29: In order to run integration tests we assumed that you have access to a Kubernetes cluster and there's the `KUBECONFIG` env set to point to the cluster. So most likely for newcomers when they run `make test` it's going to fail.

That said, I prefer keeping the tests separate and maybe update the `HACKING.md` and explain how to run unit and integration tests. WDYT?

krisctl: Thanks, @danielpacak Here is my thought process behind this change:
It was due to the scenario when I just ran the unit-tests and not integration tests locally before submitting a change and only found about it when the integration tests ran in the CI system. With test as an additional target, folks can still execute unit-tests and integration-tests separately but in addition, they also have a way to run all the tests before pulling a request.

Having that said, I am okay with simply updating `HACKING.md` to reflect the two-step operation now if that is what is preferred.

danielpacak: I see your point @krisctl . All right let's have the test target that combines all test runners. Would you update the HACKING.md and explain how to run them combined and separately and what are the prerequisites (e.g. KUBECONFIG env set)?

krisctl:
@danielpacak PTAL, thanks!

danielpacak (approved):
LGTM!

PR #137, opened 2020-08-29
Makefile error on not providing KUBECONFIG env var

### LINKED ISSUE #136: Error reporting on missing env variable could be clearer

When I run make integration-tests, I see a verbose stack trace of the errors if I don't have my pre-requisite of having KUBECONFIG env var set. 

This error reporting could be more clear if we handle this case in the Makefile itself and tell the users clearly that they are missing the env var instead of they reading through the verbose stack trace to figure that out.

**Environment:**

- Starboard version (use `starboard version`): Dev version (0.2.6 as well)
- Kubernetes version (use `kubectl version`): Any
- OS (macOS 10.15, Windows 10, Ubuntu 19.10 etc): Any

krisctl:
Fixes #136

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -35,7 +35,7 @@ unit-tests: $(SOURCES)
 
 .PHONY: integration-tests
 ## integration-tests Runs integration tests with codecov enabled.
-integration-tests: get-ginkgo
+integration-tests: check-env get-ginkgo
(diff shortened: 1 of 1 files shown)


krisctl:
@danielpacak PTAL! thanks

danielpacak (approved):
LGTM!

Issue #99, opened 2020-07-28
The cleanup command should wait for the starboard namespace termination


danielpacak:
**Describe the problem/challenge you have**

The starboard cleanup command sends a request to Kubernetes API, but does not wait for the starboard namespace termination.

For example, running `$ starboard init` immediately after `$ starboard cleanup` might fail. 

**Describe the solution you'd like**

The command should exit only if the underlying namespace is terminated

**Environment:**

- Starboard version (use `starboard version`): 0.2.6
- Kubernetes version (use `kubectl version`): Any
- OS (macOS 10.15, Windows 10, Ubuntu 19.10 etc): Any

krol3:
I'd could help with this item

PR #138, opened 2020-08-31
Issue 99 - Validate starboard ns deleted

krol3:
Related to [issue-99](https://github.com/aquasecurity/starboard/issues/99) .
Removing the Eventually in the test


pkg/kube/cr_manager.go
```diff
@@ -380,6 +381,14 @@ func (m *crManager) cleanupNamespace(ctx context.Context) (err error) {
...
 		return
 	}
+	for {
+		time.Sleep(2 * time.Second)
+		ns, err := m.clientset.CoreV1().Namespaces().Get(ctx, NamespaceStarboard, meta.GetOptions{})
+		if errors.IsNotFound(err) {
+			klog.V(3).Infof("Starboard deleted: %v", ns)
+			break
+		}
+	}
```
danielpacak, line 391:
```suggestion
	for {
		select {
		// This case controls the polling interval (every 2 seconds)
		case <-time.After(2 * time.Second):
			_, err := m.clientset.CoreV1().Namespaces().Get(ctx, NamespaceStarboard, meta.GetOptions{})
			if errors.IsNotFound(err) {
				klog.V(3).Infof("Deleted Namespace %v", NamespaceStarboard)
				return nil
			}
		// This case caters for timeout (stop polling after 30 seconds)
		case <-time.After(30 * time.Second):
			return fmt.Errorf("deleting namespace timed out")
		}
	}
```

danielpacak: The for loop with time.Sleep function would do the job as you implemented it. However, I'd suggest adding the timeout case to avoid very long polling

itest/starboard_cli_test.go
```diff
@@ -680,11 +678,6 @@ var _ = Describe("Starboard CLI", func() {
 		// TODO run by the BeforeEach callback fails when it attempts to create Kubernetes objects in the
```
danielpacak, line 678: We can remove these comments as well. They were related to deleting the namespace.

danielpacak (changes requested):
Good job @krol3 I just suggested adding check for timeout. See my comments!

pkg/kube/cr_manager.go
krol3: Thanks!! Nice suggestion !

PR #139, opened 2020-08-31
fix: Find vulnerabilities in workloads running images from private DockerHub repositories

danielpacak:

### DIFF
--- a/itest/starboard_cli_test.go
+++ b/itest/starboard_cli_test.go
@@ -2,7 +2,10 @@ package itest
 
 import (
 	"context"
+	"os"
 	"time"
+
+	"github.com/aquasecurity/starboard/pkg/kube/secrets"
 
 	"github.com/aquasecurity/starboard/pkg/apis/aquasecurity/v1alpha1"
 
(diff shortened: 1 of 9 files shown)


itest/starboard_cli_test.go
```diff
@@ -329,6 +332,112 @@ var _ = Describe("Starboard CLI", func() {
...
+			var podNamespace = corev1.NamespaceDefault
+
+			BeforeEach(func() {
+				var err error
+				var secret *corev1.Secret
+				secret, err = secrets.NewImagePullSecret(metav1.ObjectMeta{
+					Name:      secretName,
+					Namespace: podNamespace,
+				}, "https://index.docker.io/v1",
+					os.Getenv("STARBOARD_TEST_DOCKERHUB_REGISTRY_USERNAME"),
```
danielpacak, line 352: I can store registry credentials as GitHub secrets, but when we run integration tests locally I have no idea how to specify such credentials.

itest/starboard_cli_test.go
```diff
@@ -329,6 +332,112 @@ var _ = Describe("Starboard CLI", func() {
...
+						Spec: corev1.PodSpec{
+							ImagePullSecrets: []corev1.LocalObjectReference{
+								{
+									Name: secretName,
+								},
+							},
+							Containers: []corev1.Container{
+								{
+									Name:            "nginx",
+									Image:           "starboardcicd/private-nginx:1.16",
```
danielpacak, line 375: I have registered the `starbaordcicd` user in DockerHub and created the `private-nginx` repository.

pkg/kube/secret/manager.go
```diff
@@ -22,31 +23,30 @@ func NewSecretManager(clientset kubernetes.Interface) *Manager {
...
+	serverCredentials, err := s.GetRegistryHostsWithCredentials(ctx, namespace, spec.ImagePullSecrets)
 	if err != nil {
 		return
 	}
 
-	credentials = make(map[string]docker.ServerCredentials)
+	credentials = make(map[string]docker.Auth)
 	for _, image := range images {
-		server := s.GetServerFromImage(image)
+		server, err := docker.GetServerFromImageRef(image)
```
danielpacak, line 36: Here was the actual bug where imageRef for private repository was not parsed properly. For example, we returned `some-namespace` for `some-namespace/my-image:my-tag` instead of returning `index.docker.io`.

pkg/docker/config.go
```diff
@@ -3,47 +3,120 @@ package docker
...
+		return err
+	}
+	c.Auths, err = decodeAuth(c.Auths)
+	return
 }
 
-func encodeAuth(config map[string]ServerCredentials) (encodedConfig map[string]ServerCredentials, err error) {
-	encodedConfig = make(map[string]ServerCredentials)
+func decodeAuth(config map[string]Auth) (encodedConfig map[string]Auth, err error) {
+	encodedConfig = make(map[string]Auth)
```
knqyf263, line 67: It looks `decodedConfig` to me.

pkg/kube/secrets/secrets.go
```diff
@@ -0,0 +1,31 @@
...
+
+import (
+	"github.com/aquasecurity/starboard/pkg/docker"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// NewImagePullSecret constructs a new image pull Secret with the specified
+// registry server and basic authentication credentials.
+func NewImagePullSecret(meta metav1.ObjectMeta, server, username, password string) (*corev1.Secret, error) {
```
knqyf263, line 11: Is this the function for testing?

pkg/docker/config.go
knqyf263, line 67: It looks `decodedConfig` to me.

knqyf263 (commented):
Let me make sure what the PR sorted out. It used to expect a registry URL without "http/https" schema, so if `~/.docker/config.json` includes it, it didn't work. This PR adds `GetHostFromServer` and fixes the issue, right? Or, does this PR also fix something else?

danielpacak:
> Let me make sure what the PR sorted out. It used to expect a registry URL without "http/https" schema, so if `~/.docker/config.json` includes it, it didn't work. This PR adds `GetHostFromServer` and fixes the issue, right? Or, does this PR also fix something else?

Exactly that was the problem. Plus I smuggled some refactoring with better names for structs and variables.

pkg/kube/secrets/secrets.go
danielpacak: For now it's only used in the integration test, but I'll need to create a secret in the same namespace in which we create the scan Job. Currently the username/password are passed as ENVS, which has to be fixed

pkg/docker/config.go
danielpacak: Yup. Good catch. Here we decode credentials.

pkg/kube/secrets/secrets.go
knqyf263: I see.

knqyf263 (approved):

PR #141, opened 2020-09-02
fix: Set TRIVY_USERNAME and TRIVY_PASSWORD environment variables to consume secret keys

danielpacak:
So far we were passing registry credentials as values. In this PR we change it to use secretKeyRefs:

Starboard HLD - Find vulnerabilities with private image


lizrice (approved):
lgtm

PR #151, opened 2020-09-11
refactor: Init command

danielpacak:
The goal here is to review and tidy up what we're doing in the `init` command and set the stage for implementing scanners config feature:

- Renamed ServiceAccount to `starboard`
- Run all scan Jobs as `starboard` SA (+ mount API token only if necessary)
- Rename ConfigMap to `starboard` and make it generic. Introduce naming convention for keys, i.e. `trivy.` or `polaris.` to group configuration parameters by scanner. We'll introduce more config params in the future.
- Provide YAML descriptors for all objects that we create in the `init` command. This is to integrate with other tools / frameworks which do not want to or cannot execute `starboard init` command:
  ```
  k apply -f kube/init/starboard-ns.yaml \
    -f kube/init/starboard-sa.yaml \
    -f kube/init/starboard-cm.yaml \
    -f kube/init/starboard-clusterrole.yaml \
    -f kube/init/starboard-clusterrolebinding.yaml
  ```


kube/init/starboard-cm.yaml
```diff
@@ -0,0 +1,193 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  labels:
+    app.kubernetes.io/managed-by: starboard
+  name: starboard
+  namespace: starboard
+data:
```
danielpacak, line 8: This is now a generic ConfigMap. We can use it to as Volume Mount of type ConfigMap or ConfigMapKeyRef in scan Job descriptors. We can also access the values programmatically with the client-go library.

pkg/cmd/init_test.go
```diff
@@ -0,0 +1 @@
+package cmd_test
```
lizrice, line 1: is this todo?

pkg/kube/cr_manager.go
```diff
@@ -18,6 +20,11 @@ import (
 	"k8s.io/klog"
 )
 
+const (
+	clusterRoleStarboard        = "starboard"
+	clusterRoleBindingStarboard = "starboard"
+)
+
 const (
```
lizrice, line 28: Can you get rid of this `polarisConfigYAML` now there is the ConfigMap?

pkg/kube/cr_manager.go
```diff
@@ -205,74 +212,39 @@ exemptions:
...
-	if err != nil {
-		return
+	serviceAccount = &core.ServiceAccount{
+		ObjectMeta: meta.ObjectMeta{
+			Name: ServiceAccountStarboard,
+			Labels: labels.Set{
+				"app.kubernetes.io/managed-by": "starboard",
+			},
+		},
+		AutomountServiceAccountToken: pointer.BoolPtr(false),
```
lizrice, line 231: isn't this defined in the YAML?

pkg/kube/cr_manager.go
```diff
@@ -205,74 +212,39 @@ exemptions:
...
+		ObjectMeta: meta.ObjectMeta{
+			Name: ConfigMapStarboard,
+			Labels: labels.Set{
+				"app.kubernetes.io/managed-by": "starboard",
+			},
+		},
+		Data: map[string]string{
+			"trivy.severity":      "UNKNOWN,LOW,MEDIUM,HIGH,CRITICAL",
+			"polaris.config.yaml": polarisConfigYAML,
+		},
```
lizrice, line 243: Is this needed if it's defined in starboard-cm.yaml?

lizrice (commented):
I'm not sure, but I think you can delete some of the literals in the code in favour of the new ConfigMap?

pkg/cmd/init_test.go
danielpacak: This file is required to properly measure test coverage. However, it does't make much sense to test the init.cmd code with unit tests because we'll end up mocking Cobra library. So this is to generate coverage report for integration tests runner, which eventually is merged with unit tests runner's report.

pkg/kubebench/scanner.go
```diff
@@ -132,9 +132,11 @@ func (s *Scanner) prepareKubeBenchJob(node core.Node) *batch.Job {
 					},
 				},
 				Spec: core.PodSpec{
-					RestartPolicy: core.RestartPolicyNever,
-					HostPID:       true,
-					NodeName:      node.Name,
+					ServiceAccountName:           kube.ServiceAccountStarboard,
+					AutomountServiceAccountToken: pointer.BoolPtr(true),
```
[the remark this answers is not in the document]
danielpacak: We have this flags set to false on ServiceAccount level. For kube-bench and Polaris, which access Kubernetes API, we do want to mount the service token. It's not required for Trivy and kube-hunter though.

pkg/kube/cr_manager.go
danielpacak: I'd love to get rid of duplicate struct and string literals in favour of YAML files embedded in starboard executable (see https://github.com/golang/go/issues/41191) So for now the question is whether we want to get rid of starboard init command and use kubectly apply as the init step. I thought that these are alternative ways to initialize starboard. WDYT?

lizrice: oh, I see... I think `starboard init` is clearer and easier than having to manually do an `apply`

lizrice (approved):

Issue #144, opened 2020-09-08
Non-standard syntax for Kubernetes labels


danielpacak:
According to https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set labels have the following syntax: `<prefix>/<name>`, where `/kubernetes.io` and `k8s.io/` prefixes are reserved for Kubernetes core components.

To follow this convention we should define label prefix for Starboard, e.g.: `aquasecurity.github.io/` or `starboard.github.io/` or `starboard.aquasecurity.github.io/`

Currently used labels:
- starboard.container.name=nginx
- starboard.resource.kind=Deployment
- starboard.resource.name=nginx
- starboard.resource.namespace=default

Should be:
- aquasecurity.github.io/starboard-container-name=nginx
- aquasecurity.github.io/starboard-resource-kind=Deployment
- aquasecurity.github.io/starboard-resource-name=nginx
- aquasecurity.github.io/starboard-resource-namespace=default

danielpacak:
We don't do that for now as I'm not convinced to request users to type in such long labels in the command line.

PR #153, opened 2020-09-16
refactor: Use standard syntax for Kubernetes labels

danielpacak:
Resolves: #144


simar7:
If you would like to run codecov on each PR but not indicate a failure in case coverage drops (or if you want to set a certain threshold of allowed drop) you can do something like this https://github.com/aquasecurity/trivy/blob/master/codecov.yml#L1-L12

I personally like to only fail on situation when something bad happens, coverage drop of a `0.20%` is I feel insignificant. Just a suggestion.

pkg/cmd/get_configaudit.go
```diff
@@ -43,10 +46,16 @@ NAME is the name of a particular Kubernetes workload.
...
+				kube.LabelResourceKind:      string(workload.Kind),
+				kube.LabelResourceName:      workload.Name,
+				kube.LabelResourceNamespace: workload.Namespace,
+			}.String()
+
 			kubectlCmd := exec.Command("kubectl",
 				"get",
 				starboard.ConfigAuditReportCRName,
-				fmt.Sprintf("-l=starboard.resource.kind=%s,starboard.resource.name=%s", workload.Kind, workload.Name),
+				fmt.Sprintf("-l=%s", selector),
```
simar7, line 58: So functionally it returns the same info by doing a `.String()` on a `labels.Set()` call? Is there a test I could read up on?

PR #157, opened 2020-09-21
Added Last Updated Field For Reports

### LINKED ISSUE #150: GeneratedAt property of VulnerabilityReport is not displayed

Generate HTML report for a Deployment and open it in a Web browser:

starboard_html_report_generated_at_blank

The Generated At value is blank

I'd expect that the Generated At is displayed as formatted date time.

N/A

**Environment:**

- Starboard version (use `starboard version`): 0.3.0
- Kubernetes version (use `kubectl version`): Any
- OS (macOS 10.15, Windows 10, Ubuntu 19.10 etc): Any

danielsagi:
HTML report was missing a correct LastUpdated timestamp.
The old way we used to get this timestamp is by looking at the Creation Timestamp: [Example from octant-plugin](https://github.com/aquasecurity/starboard-octant-plugin/blob/2655c5b5c00c6ea9f86858f0f55e3865b3d1787b/pkg/plugin/view/configaudit/report_view.go#L46)
This method was also used in the html report so far. (for ConfigAudits)

### The Problem
Today reports are created with a "Create Or Update" logic. 
So Creation Timestamp of the kubernetes resource is irrelevant. 

### Solution
Starting from kuberentes 1.18 we have "Managed Fields" which can contain the last `Update` Timestamp of the resource.
However, we want to support older versions as well, The best way is adding this field to the report itself. So we could easily access it.
This way is also mentioned [in](https://github.com/goharbor/pluggable-scanner-spec/blob/5f4f631765ce7ed23b91f138407c2b954831536b/api/spec/scanner-adapter-openapi-v1.0.yaml#L294)

This PR introduces:
* A new LastUpdated field in following reports 
  * ConfigAudits 
  * Vulneratbilities
  * CisBenchMark
  * KubeHunter
* Usage of the new field in HTML reports generations

closes #150


pkg/apis/aquasecurity/v1alpha1/common_types.go
```diff
@@ -1,8 +1,14 @@
...
+import "time"
+
 // Scanner is the spec for a scanner generating a security assessment report.
 type Scanner struct {
 	Name    string `json:"name"`
 	Vendor  string `json:"vendor"`
 	Version string `json:"version"`
 }
+
+type GeneratedAt struct {
```
danielpacak, line 12: Hey @danielsagi This is a tricky one. I should have put more details into the issue description:

Each Kubernetes object, including custom object such as VulnerabilityReport and ConfigAuditReport, has `CreationTimestamp` ObjectMeta property that can be used instead of defining our own. We also use the CreationTimestamp property in Octant plugin.

We actually created a similar thing to custom GeneratedAt property long time back to make sure this date is independent of the etcd system date, but it has its own drawbacks. For example, if you want to calculate content hash of a VulnerabilityReport to optimize storage, the GeneratedAt property would change the hash.

danielsagi: Yes I know CreationTimestamp is in ObjectMeta, but there is no LastUpdatedTimestamp.
For example if we run multiple scans on the same deployment, the CreationTimestamp stays the same. as the resource was created firstly, and then updated.

pkg/apis/aquasecurity/v1alpha1/config_audit_types.go
```diff
@@ -70,6 +70,7 @@ type ConfigAuditReportList struct {
 // TODO my-node)
 type ConfigAudit struct {
 	Scanner         Scanner            `json:"scanner"`
+	LastUpdated     LastUpdated        `json:"lastUpdated"`
```
danielpacak, line 73: In `ObjectMata` there's the `CreationTimestamp` field so maybe we should call this one `UpdateTimestamp`. I don't have strong opinions on that. I just like consistency.

danielpacak (changes requested):

danielpacak (approved):

Issue #55, opened 2020-06-13
Output init manifests


billyshambrook:
**Describe the problem/challenge you have**

To install starboard via GitOps, we need the YAML manifests to install. Would be great if the init command could optionally just output the manifests to stdout.

**Describe the solution you'd like**

**Anything else you would like to add:**

Happy to help with a PR if this sounds good and there's a preferable cmd line argument to trigger this (e.g. `starboard manifests`)

**Environment:**

- Starboard version (use `starboard version`):
- Kubernetes version (use `kubectl version`):
- OS (macOS 10.15, Windows 10, Ubuntu 19.10 etc):

danielpacak:
Thank you @billyshambrook for bringing the GitOps use case. I think it's reasonable to allow getting the manifests in the init command as you described. To some extent it's also related to #23 

I'd suggest adding the `--dry-run` and `--output` flags to the existing init command:

```
$ starboard init --dry-run -o yaml
```

We want to stay as close to the kubectl syntax as possible, so the above mentioned command would resemble a build-in kubectl commands:

```
 $ k create deploy nginx --image nginx:1.16 --dry-run -o yaml
```

WDYT? If that sound's good I'm happy to assign you @billyshambrook to this one.

shaunmclernon:
@danielpacak I have been digging around the code on this one.

Based on the brief above, I am planning on possibly breaking this up into 2 PRs (1 for the dry-run and the other for yaml generation). I propose to mimic the kubectl behaviour for dry-run to show which objects are to be created (it is sort of an extension to work done in #164)

Currently, I have the following resources working in dry-run mode:

> $ starboard init --dry-run
Creating CRD "vulnerabilityreports.aquasecurity.github.io" (dry run)
Creating CRD "ciskubebenchreports.aquasecurity.github.io" (dry run)
Creating CRD "kubehunterreports.aquasecurity.github.io" (dry run)
Creating CRD "configauditreports.aquasecurity.github.io" (dry run)
Creating Namespace "starboard" (dry run)
Creating ConfigMap "starboard/starboard" (dry run)

Investigating the service account, cluster role and cluster role binding (and the relationship between #23)  

and obviously the YAML objects when -o yaml is included.

WDYT? Does that sound like a reasonable approach?  If so I am happy for you to assign it to me.

danielpacak:
> @danielpacak I have been digging around the code on this one.
> 
[...]

PR #196, opened 2020-10-12
feat: Add dry-run option to init command

shaunmclernon:
Added an additional flag to the init command to allow it run as a
dry-run. Where ever possible, it follows the same convention as kubectl.

Example:

$ starboard init --dry-run

crd/vulnerabilityreports.aquasecurity.github.io created (dry run)
crd/ciskubebenchreports.aquasecurity.github.io created (dry run)
crd/kubehunterreports.aquasecurity.github.io created (dry run)
crd/configauditreports.aquasecurity.github.io created (dry run)
namespace/starboard created (dry run)
configmap/starboard created (dry run)
serviceaccount/starboard created (dry run)
clusterrole.rbac.authorization.k8s.io/starboard created (dry run)
clusterrolebinding.rbac.authorization.k8s.io/starboard created (dry run)

The expection is that when running this command no resources should be
created in the cluster. Integrations tests have been added to verify this.

This is part of #55 but does not resolve it completely.

Signed-off-by: Shaun McLernon <shaun@codesome.tech>

### DIFF
--- a/itest/starboard_cli_test.go
+++ b/itest/starboard_cli_test.go
@@ -64,6 +64,100 @@ func NewPodSpec(podName string, containers map[string]string, args ...string) (*
 	return pod, err
 }
 
+var _ = Describe("Starboard CLI dryrun", func() {
+
+	Describe("Command init with dryrun", func() {
+
+		It("should return an error if the wrong dryrun flag is provided", func() {
+			out := NewBuffer()
+			err := cmd.Run(versionInfo, []string{
+				"starboard",
+				"init",
+				"--dry",
+				"-v", starboardCLILogLevel,
+			}, out, out)
+			Expect(err).To(HaveOccurred())
+
+			// Expectation: No resources should have been created, so lets check
+
+			crdList, err := customResourceDefinitions.List(context.TODO(), metav1.ListOptions{})
+			Expect(err).ToNot(HaveOccurred())
+			Expect(crdList.Items).Should(HaveLen(0))
+		})
+
+		It("should return an error if dryrun flag provides invalid parameters", func() {
+			out := NewBuffer()
+			err := cmd.Run(versionInfo, []string{
+				"starboard",
+				"init",
+				"--dry-run=client",
+				"-v", starboardCLILogLevel,
+			}, out, out)
+			Expect(err).To(HaveOccurred())
+
+			// Expectation: No resources should have been created, so lets check
+
+			crdList, err := customResourceDefinitions.List(context.TODO(), metav1.ListOptions{})
+			Expect(err).ToNot(HaveOccurred())
+			Expect(crdList.Items).Should(HaveLen(0))
+		})
+
+		It("should show which resources will be created", func() {
+			out := NewBuffer()
+			err := cmd.Run(versionInfo, []string{
+				"starboard",
+				"init",
+				"--dry-run",
+				"-v", starboardCLILogLevel,
+			}, out, out)
+			Expect(err).ToNot(HaveOccurred())
+
+			// Expectation: No resources should have been created, so lets check
+
+			Eventually(out).Should(Say(`crd/vulnerabilityreports.aquasecurity.github.io created \(dry run\)`))
+			Eventually(out).Should(Say(`crd/ciskubebenchreports.aquasecurity.github.io created \(dry run\)`))
+			Eventually(out).Should(Say(`crd/kubehunterreports.aquasecurity.github.io created \(dry run\)`))
+			Eventually(out).Should(Say(`crd/configauditreports.aquasecurity.github.io created \(dry run\)`))
+			crdList, err := customResourceDefinitions.List(context.TODO(), metav1.ListOptions{LabelSelector: "app.kubernetes.io/managed-by=starboard"})
+			Expect(err).ToNot(HaveOccurred())
+			Expect(crdList.Items).Should(HaveLen(0))
+
+			Eventually(out).Should(Say(`namespace/starboard created \(dry run\)`))
+			Eventually(out).Should(Say(`configmap/starboard created \(dry run\)`))
+			cm, err := configmaps.List(context.TODO(), metav1.ListOptions{})
+			Expect(err).ToNot(HaveOccurred())
+			Expect(cm.Items).Should(HaveLen(0))
+
+			Eventually(out).Should(Say(`serviceaccount/starboard created \(dry run\)`))
+			sa, err := serviceAccounts.List(context.TODO(), metav1.ListOptions{LabelSelector: "app.kubernetes.io/managed-by=starboard"})
+			Expect(err).ToNot(HaveOccurred())
+			Expect(sa.Items).Should(HaveLen(0))
+
+			Eventually(out).Should(Say(`clusterrole.rbac.authorization.k8s.io/starboard created \(dry run\)`))
+			cr, err := clusterRoles.List(context.TODO(), metav1.ListOptions{LabelSelector: "app.kubernetes.io/managed-by=starboard"})
+			Expect(err).ToNot(HaveOccurred())
+			Expect(cr.Items).Should(HaveLen(0))
+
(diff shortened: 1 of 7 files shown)


shaunmclernon:
@danielpacak sorry, this took me way longer than I thought, mainly due to the integration testing.

There is an issue that we cannot do a "real" dry-run for the resources in the starboard namespace. I have flagged them with a FIXME to make it easier to find.  I did have a look a the kubectl implementation however, it is slightly more complicated than here.

itest/starboard_cli_test.go
```diff
@@ -64,6 +64,91 @@ func NewPodSpec(podName string, containers map[string]string, args ...string) (*
 	return pod, err
 }
 
+var _ = Describe("Starboard CLI dryrun", func() {
```
danielpacak, line 67: I believe that we should keep **Starboard CLI** as top level Description and add these new `It` cases under **Starboard CLI / Command init**. There's already `It("should initialize Starboard")` and you're exercising it with the --dry-run flag so we'd better keep the tests together.

So our goal is that all specs related to the init command read as follows:
- Starboard CLI / Command init / should initialize Starboard
- Starboard CLI / Command init / should return an error if the wrong dryrun flag is provided
- Starboard CLI / Command init / should return an error if dryrun flag provides invalid parameters
- Starboard CLI / Command init /should show which resources will be created

itest/starboard_cli_test.go
```diff
@@ -64,6 +64,91 @@ func NewPodSpec(podName string, containers map[string]string, args ...string) (*
...
+
+		It("should return an error if the wrong dryrun flag is provided", func() {
+			out := NewBuffer()
+			err := cmd.Run(versionInfo, []string{
+				"starboard",
+				"init",
+				"--dry",
+				"-v", starboardCLILogLevel,
+			}, out, out)
+			Expect(err).To(HaveOccurred())
```
danielpacak, line 79: Did you mean testing a wrong flag (--dry instead of --dry-run) or the wrong value of the --dry-run flag?

I think in this case we'll get the following error: `error: unknown flag: --dry`

itest/starboard_cli_test.go
```diff
@@ -64,6 +64,91 @@ func NewPodSpec(podName string, containers map[string]string, args ...string) (*
...
+
+			// Expectation: No resources should have been created, so lets check
+
+			crdList, err := customResourceDefinitions.List(context.TODO(), metav1.ListOptions{})
+			Expect(err).ToNot(HaveOccurred())
+			Expect(crdList.Items).Should(HaveLen(0))
+		})
+
+		It("should show which resources will be created", func() {
+			out := NewBuffer()
```
danielpacak, line 106: Can we assert on this output as well? See how we check the output for the version command.

pkg/cmd/init.go
```diff
@@ -43,9 +44,12 @@ These resources can be removed from the cluster using the "cleanup" command.
 			if err != nil {
 				return
 			}
-			err = kube.NewCRManager(clientset, clientsetext).Init(ctx)
+
+			err = starboard.NewCRManager(clientset, clientsetext).Init(ctx, options, outWriter)
 			return
 		},
 	}
+	cmd.Flags().BoolVar(&options.DryRun, "dry-run", false, "Only print the object that would be sent, without sending it.")
```
danielpacak, line 52: If I look at the output of `kubectl create deployment -h` the --dry-run flag accepts 3 values. In our case it's a boolean. Can we make it consistent?

--dry-run='none': Must be "none", "server", or "client". If client strategy, only print the object that would be
sent, without sending it. If server strategy, submit server-side request without persisting the resource.

pkg/kube/cr_manager.go
```diff
@@ -148,54 +149,56 @@ func NewCRManager(clientset kubernetes.Interface, clientsetext extapi.Apiextensi
 	}
 }
 
-func (m *crManager) Init(ctx context.Context) (err error) {
-	err = m.createOrUpdateCRD(ctx, &aquasecurityv1alpha1.VulnerabilityReportsCRD)
+func (m *crManager) Init(ctx context.Context, initOptions *InitOptions, outWriter io.Writer) (err error) {
```
danielpacak, line 152: If you passed the initOptions and the output Writer as args to the NewCRManager constructor function and saved it as crManager properties you'd not need to change methods' signatures in so many places.

pkg/kube/cr_manager.go
```diff
@@ -148,54 +149,56 @@ func NewCRManager(clientset kubernetes.Interface, clientsetext extapi.Apiextensi
...
+	klog.V(3).Infof("In DryRun mode=%v", initOptions.DryRun)
```
danielpacak, line 153: I think we could simplify this logic and resolve --dry-run mode in one place instead of spreading this logic in so many places.

I'd actually start with checking if dry-run mode is `client` and immediately short circuit the `Init` method with print statements showing which objects will be created without sending any requests to Kubernetes API.

Then we have to handle `none` and `server` modes, but this I'd even resolve in the construction and store CreateOptions and UpeateOptions structs as crManager fields, i.e.:

```go
type crManager struct {
	clientset    kubernetes.Interface
	clientsetext extapi.ApiextensionsV1beta1Interface
	createOptions meta.CreateOptions
	updateOptions meta.UpdateOptions
}

// NewCRManager constructs a CRManager with the given Kubernetes interface.
func NewCRManager(opt InitOptions, clientset kubernetes.Interface, clientsetext extapi.ApiextensionsV1beta1Interface) CRManager {
	return &crManager{
		clientset:    clientset,
		clientsetext: clientsetext,
		createOptions: meta.CreateOptions{
			// Set DryRun property based on the opt passed as constructor arg
			DryRun: []string{},               // if --dry-run=none
			DryRun: []string{meta.DryRunAll}, // if --dry-run=server
		},
		updateOptions: meta.UpdateOptions{
			DryRun: []string{},
			DryRun: []string{meta.DryRunAll},
		},
	}
}
```

WDYT?

danielpacak (changes requested):
I know this is not easy @shaunmclernon but you did great progress anyway. Please see my comments and let me know if it makes sense.

itest/starboard_cli_test.go
shaunmclernon: Yes, I was testing an invalid option... which should return an error as you have noted.

pkg/kube/cr_manager.go
shaunmclernon: Thanks for that... I was worried that I was making a lot of signature changes.

itest/starboard_cli_test.go
shaunmclernon: That was my plan, although I wasn't able to achieve that because there was a before step that was performing the "init". Although, it is probably my lack of ginkgo knowledge... I will have another crack at it.

pkg/kube/cr_manager.go
shaunmclernon: Now that I have been through the process and know how it works... then I would be in favour of simplifying the logic. Your approach sounds sensible, so I will refactor accordingly.

itest/starboard_cli_test.go
shaunmclernon: Done

itest/starboard_cli_test.go
danielpacak: Ah I see. If it's getting too complex I'm okey with your initial implementation... I forgot that we call init and cleanup command around each test case ...

itest/starboard_cli_test.go
shaunmclernon: Sure, I have added this now.

danielpacak:
@shaunmclernon Are you still working on this one?

shaunmclernon:
@danielpacak Yes, sorry it's taking so long.

danielpacak:
> @danielpacak Yes, sorry it's taking so long.

Sure. No rush. Just let me know if you need any help.

danielpacak:
@shaunmclernon is this PR something you are working on? Otherwise I believe we can close it.

danielpacak:
Closing as stale pull request

PR #201, opened 2020-10-14
docs: Add operator to the README.md

danielpacak:


README.md
```diff
@@ -249,32 +288,73 @@ The following table lists available configuration parameters.
...
+| --------------- | ------------------ | -------------------------- | ----------- |
+| OwnNamespace    | `operators`        | `operators`                | The operator can be configured to watch events in the namespace it is deployed in. |
+| SingleNamespace | `operators`        | `foo`                      | The operator can be configured to watch for events in a single namespace that the operator is not deployed in. |
+| MultiNamespace  | `operators`        | `foo,bar,baz`              | The operator can be configured to watch for events in more than one namespace. |
+| AllNamespaces   | `operators`        |                            | The operator can be configured to watch for events in all namespaces. |
+
+> **CAUTION:** Although we do support the *AllNamespaces* install mode, please use it with caution when your cluster
+> runs a moderate or high number of workloads. If the desired state of the cluster is much different from the actual
+> state, the operator might spin up too many scan jobs and negatively impact the performance of your cluster.
+> We're working on improvements to limit the number of parallel scan jobs and implement a back pressure logic.
```
lizrice, line 341:
```suggestion
> We're planning improvements to limit the number of parallel scan jobs and implement a back pressure logic.
```

README.md
```diff
@@ -215,7 +226,35 @@ vulnerabilities as well as configuration issues that might affect stability, rel
...
+[custom security resources][starboard-crds].
+
+To learn more about the available Starboard CLI commands, run `starboard help` or type a command followed by the
+`-h` flag:
+
+```
+$ starboard kube-hunter -h
+```
+
+### Configuration
```
lizrice, line 257: Am I right to think that the Starboard operator will create the vulnerabilityreports CRD - no need to run `starboard init` first?

lizrice (commented):
I'm wondering what we should do with the Installation, Getting Started & Next Steps sections. 

- I think it makes sense to Get Started with the command line tool and then move on to trying the operator when you see what Starboard is all about. Maybe we should explicitly suggest that? 
- We need installation instructions for the operator, I think?

README.md
danielpacak: It depends how we install the operator. With OLM the OLM operator is taking care of creating CRDs, which are part of the installation bundle. For helm or kubecly apply installations we have to create the CRDs manually or use the init command.

In general, as a good practice the operator should not create the CRDs that it manages.

README.md
```diff
@@ -290,6 +294,48 @@ security resources as depicted below. However, we plan to support all [custom se
...
+
+   ```
+   $ kubectl apply -f deploy/static/05-starboard-operator.deployment.yaml
+   ```
+
+#### With Helm
+
+Helm, which is de facto standard package manager for Kubernetes, allows installing applications from parameterized
+YAML manifests called Helm charts. To address shortcomings of static YAML manifest we'll provide a Helm chart to deploy
+the Starboard operator. See [#187](https://github.com/aquasecurity/starboard/issues/187) to check the progress on that.
```
consideRatio: I suggest to let the static resources be generated from the Helm chart in the future. See #211

danielpacak: I totally agree @consideRatio Once we merge the Helm chart, we'll remove static manifests from `./deploy/static/*.yaml` and update the docs to use the `helm template` command for those who cannot use Helm CLI

README.md
consideRatio: Helm will install them if they dont already excist now, but not as the other resources. It will also not try to update them or similarly if, just warn about it if needed.

README.md
```diff
@@ -249,6 +283,182 @@ The following table lists available configuration parameters.
...
+have to create new manifests or edit existing ones.
+
+To deploy the operator in the `starboard-operator` namespace and configure it to watch the `default`
+namespace:
+
+1. Send the definition of the [vulnerabilityreports][vulnerabilityreports-crd] custom resource to the Kubernetes API:
+
+   ```
+   $ kubectl apply -f deploy/crd/vulnerabilityreports.crd.yaml
+   ```
```
consideRatio, line 317: I think this step wont be needed.

README.md
```diff
@@ -249,6 +283,182 @@ The following table lists available configuration parameters.
...
+parameterized YAML manifests called Helm [charts][helm-charts].
+
+To address shortcomings of static YAML manifests we provide the Helm chart to deploy the Starboard operator. The
+[starboard-operator](./deploy/helm) Helm chart supports all [install modes](#install-modes). For example, to install
+the operator in the `starboard-operator` namespace and configure it to watch `foo` and `bar` namespaces, run:
+
+```
+$ kubectl create ns starboard-operator
+$ helm install starboard-operator -n starboard-operator \
+    ./deploy/helm \
```
consideRatio, line 344: There is also the --create-namespace flag btw

danielpacak: Great. I didn't know that. Thanks for letting us know. I used the flag and merged the two commands into one!

README.md
danielpacak: This is great that a Helm chart package now supports CRDs that are installed before a Helm release is created. Even though it's not as powerful and automated as OLM, it's useful in case of Helm chart installation option

README.md
```diff
@@ -36,71 +43,27 @@ relate to different resources in a Kubernetes-native way. Starboard provides
...
-```
+![](docs/images/starboard-cli-with-octant-demo.gif)
 
 ## Getting Started
 
 The easiest way to get started with Starboard is to use [Starboard CLI][starboard-cli], which allows scanning Kubernetes
 workloads deployed in your cluster.
 
+> **NOTE:** Even though manual scanning through the command-line is useful, it doesn't scale well with a huge number
+> of Kubernetes workloads and / or multi-tenant clusters. Therefore, the [Starboard Operator][starboard-operator]
```
lizrice, line 63:
```suggestion
> **NOTE:** Even though manual scanning through the command-line is useful, the fact that it's not automated makes it less suitable with a large number
> of Kubernetes workloads. Therefore, the [Starboard Operator][starboard-operator]
```

I'm assuming that automation is the main motivator here - my suggestion is to avoid the possible interpretation that "this is written badly so even installing it it might cause problems in your large production cluster", which I don't think is the case

lizrice (approved):
I made one suggestion but LGTM

PR #197, opened 2020-10-13
feat: helm chart

krol3:
Issue [187](https://github.com/aquasecurity/starboard/issues/187) - create helm chart for starboard

PR #204, opened 2020-10-14
[operator] Add Helm chart

### LINKED ISSUE #187: [operator] Allow deploying operator with Helm chart
Some folks might not want to install OLM and Operator Marketplace in their (vanilla) Kubernetes clusters. In such cases we can provide a Helm chart and supports all install modes as defined by OLM:

- OwnNamespace
- SingleNamespace
- MultiNamespace
- AllNamespaces

The Helm chart would also take care of creating RBAC objects based on `operatorNamespace` and `operatorTargetNamespaces` values.

consideRatio:
I understood that you are accepting contributions where a Helm chart is defined for the starboard-operator, so this is meant to fix #187. Me and @krol3 both started working on this in parallel, but as discussed in https://github.com/aquasecurity/starboard/pull/197#issuecomment-708356202 continue with a PR here that @krol3 will review!

## PR ambition
In this PR, I've tried to follow the current state of the evolving best practices of Helm charts and created a foundation that will be relatively easy to maintain in the future. As an example, I've tried to avoid the anti-patterns of hardcoding support for a limited set of environment variables which would have needed to be updated as the starboard-operator binary added more configuration options.

## Part of this PR
- [x] Create a Helm chart to install the starboard-operator
- [x] Expose the /metrics endpoint and make it easy for prometheus to scrape it.
- [x] Provide a message in NOTES.txt to show after `helm install|upgrade`
- [x] Verify functionality in a k8s cluster in all its operational states.
  - Activating it cluster wide led to ~70 pods created -> #209
  - Scanner jobs didn't cache the vulnerability database, and 403 rate limit errors from GitHub API followed.
  - Some jobs ended up resulting in such a large VulnerabilityReport that registering it with the k8s api-server was refused -> #208
  - Can I configure the pods created by starboard-operator, or specifically, can I provide them with a github token for trivy to use in order to avoid being rate limited by GitHub? -> #210

## Undecided if part of PR
- [ ] Create a GitHub Actions CI test running `helm template` which only triggers when the chart folder is changed.
- [ ] Document the experimental state of the Helm chart somewhere

## Not part of this PR
- Create a Helm chart registry where this Helm chart be published
- Define CI jobs to package and publish the Helm chart to a Helm chart registry
- Define CI jobs to start a local k8s cluster, install it, and verify it functions
- Create a `values.schema.json` that automatically validates passed Helm values to `helm template|install|upgrade`

## Things to consider
- [x] Are configuration options named sensibly?
- [x] Are configuration options documented in values.yaml good enough?
- [x] I've assumed that starboard-operator does not support being run alongside another starboard-operator, and that it wouldn't make sense to require it to be running in a highly available configuration. Is this correct?


consideRatio:
@krol3 I consider this ready for review at this point! I'm running low on time to do everything I planned (basic CI test for example), but this is now in a functional state it seems from my local testing.

deploy/helm/templates/rbac.yaml
```diff
@@ -0,0 +1,82 @@
...
+{{- if .Values.rbac.create }}
+---
+{{- /*
+Create (Cluster)Role and (Cluster)RoleBinding depending on if the Helm chart is
+installed in a namespace different from the targetNamespace.
+*/}}
+{{- $clusterWide := not (eq .Release.Namespace (tpl .Values.targetNamespaces .)) }}
+{{- $conditionalClusterPrefix := $clusterWide | ternary "Cluster" "" }}
+apiVersion: rbac.authorization.k8s.io/v1
+kind: {{ $conditionalClusterPrefix }}Role
```
consideRatio, line 24: @danielpacak you wrote the following in #187:

> The Helm chart would also take care of creating RBAC objects based on operatorNamespace and operatorTargetNamespaces values.

1. I've made the `operatorNamespace` be in whatever namespace the Helm chart is installed, so no such configuration is exposed.
2. I've made `targetNamespaces` be a configuration of the Helm chart, which is by default the same as the namespace the Helm chart is installed.
3. The logic to decide of (Cluster)Role (Cluster)RoleBinding should be installed can be inspected above.

Does this seem like an okay implementation?

krol3:
Hi @consideRatio I tested and It's working the helm chart. I will close my PR.

deploy/helm/templates/deployment.yaml
```diff
@@ -0,0 +1,107 @@
...
+      containers:
+        - name: "{{ .Chart.Name }}"
+          image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
+          env:
+            - name: OPERATOR_NAMESPACE
+              value: {{ .Release.Namespace | quote }}
+            - name: OPERATOR_TARGET_NAMESPACES
+              value: {{ tpl .Values.targetNamespaces . | quote }}
+            - name: OPERATOR_METRICS_BIND_ADDRESS
+              value: ":8080"
```
krol3, line 38: We could use: {{ print ":" .Values.image.metricsPort | quote }}

deploy/helm/templates/deployment.yaml
```diff
@@ -0,0 +1,107 @@
...
+                  key: {{ $key | quote }}
+                  optional: true
+            {{- end }}
+            {{- end }}
+          ports:
+            # metrics port is exposed by the k8s Service and by default also
+            # have annotations which will help prometheus as a target for
+            # scraping of metrics
+            - name: metrics
+              containerPort: 8080
```
krol3, line 57: Also here: {{ .Values.image.metricsPort }}

deploy/helm/templates/deployment.yaml
consideRatio: I think we can avoid needing to add configuration of this because users will only access this through the k8s Service, which in turn points to the Pod's port named `metrics`.

Hmmm, the k8s Service port can currently be configured with `service.port`, but I think we should name that `metricsPort` like you suggest here as that probably makes it less confusing about what the service is meant for currently, and also in case we want to expose something else that isn't metrics on another port.

consideRatio: Users of another Helm chat I've worked with have been fine without configuring the container/pod port for a very long time, so we opted there to not add it at any point in time: https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/master/jupyterhub/templates/hub/deployment.yaml#L199-L201

consideRatio:
Thank you @krol3 for your review and testing that things seem to work on your end as well :heart: :tada:

deploy/helm/templates/deployment.yaml
consideRatio: I added 9143ca1 to rename service.port to service.metricsPort - do you agree this is sufficient?

deploy/helm/values.yaml
```diff
@@ -0,0 +1,101 @@
...
+  # name specifies the name of the k8s Secret to reference. If not set, a name
+  # is generated using the fullname template. name is currently required to be
+  # hardcoded to starboard-operator as the starboard-operator binary makes a
+  # hardcoded reference to that name:
+  # https://github.com/aquasecurity/starboard/blob/ad6b9b99931d7524806f79b1329fdda82c75a715/pkg/operator/scanner.go#L26-L28
+  name: "starboard-operator"
+  # stringData specifies key value pairs to be added to the k8s Secret and
+  # mounted by the Pod as environment variables.
+  #
+  # NOTE: It is important that you specify targetNamespace instead of
```
danielpacak, line 47: To complete that node, maybe we could mention that also the OPEARTOR_NAMESPACE env is inferred from a Helm release namespace. WDYT?

deploy/helm/values.yaml
```diff
@@ -0,0 +1,101 @@
...
+    OPERATOR_SCANNER_TRIVY_ENABLED: "true"
+    OPERATOR_SCANNER_TRIVY_VERSION: "0.12.0"    # https://hub.docker.com/r/aquasec/trivy/tags
+    OPERATOR_SCANNER_AQUA_CSP_ENABLED: "false"
+    OPERATOR_SCANNER_AQUA_CSP_VERSION: ""
+    OPERATOR_SCANNER_AQUA_CSP_HOST: ""
+    OPERATOR_SCANNER_AQUA_CSP_USER: ""
+    OPERATOR_SCANNER_AQUA_CSP_PASSWORD: ""
+
+rbac:
+  create: true
```
danielpacak, line 61: What' would be a legit use case to set this value to `false`? When someone granted cluster-admin permission to all service accounts via `service:serviceaccounts` group?

danielpacak (commented):
Great job @consideRatio ! I really appreciate your contribution. I also tested the Helm chart in my cluster and it works as expected.

I left a few comments / questions and in general I do agree with your assumptions that the follow up tasks, such as CI integration and updates to the README, can be done in separate PRs.

Before we merge this one, I'm wondering whether we should add a template to define the VulnerabilityReports custom resource. Some people say that the operator itself should not install the CRDs that it manages, but here we have a Helm chart, which provides means of installation.

Beyond that, I assume that the Helm chart is self contained, i.e. does not require `$ starboard init` or `$ kubectl starboard init` command to be run.

That said, do you think we can somehow symbolically link to https://github.com/aquasecurity/starboard/blob/master/deploy/crd/vulnerabilityreports.crd.yaml and send it to Kubernetes API along with other Helm templates? Otherwise we may assume that a deployer defines CRs with `kubectl create` command.

I also realized that the operator should (programmatically) check if CRs are defined before it spawns any scan job. Otherwise we're waisting resources just to find out that we cannot save a report because of an unknown resource.

deploy/helm/values.yaml
consideRatio: When the account that deploys on regular basis doesnt have permissions to setup clusterroles for example i think.

I'm uncertain id this is a practice that remains helpful for some people still though, mostly added it because i know people have requested these aspects in the past for other helm charts.

deploy/helm/values.yaml
consideRatio: Good point, commit added!

deploy/helm/values.yaml
```diff
@@ -0,0 +1,101 @@
...
+
+podAnnotations: {}
+
+podSecurityContext: {}
+  # fsGroup: 2000
+
+securityContext:
+  runAsNonRoot: true
+  allowPrivilegeEscalation: false
+  runAsUser: 65534
```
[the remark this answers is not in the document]
consideRatio: Addressed with a commit.

consideRatio:
Thank you for your review @danielpacak! :tada: :heart:

> Before we merge this one, I'm wondering whether we should add a template to define the VulnerabilityReports custom resource. Some people say that the operator itself should not install the CRDs that it manages, but here we have a Helm chart, which provides means of installation.

Helm 3 supports this, but it is an evolving best practice. I think its the right call to bundle the CRDs with the helm chart. They won't be templates that render with values, and they won't be managed by Helm after install in general I think. For more info, see: https://helm.sh/docs/chart_best_practices/custom_resource_definitions/

- [x] CRDs added
- [x] container securityContext runAsUser/runAsGroup updated to match predefined user/group in Dockerfile
- [x] Note about OPERATOR_NAMESPACE being inferred added.
- [ ] Decide if `rbac.create: true` is to remain or be hardcoded as true. I'm not confident about the matter myself, but I suggest to let it remain unless better understanding about why so many helm chart has this setup and its no longer relevant for this helm chart.

deploy/helm/values.yaml
danielpacak: Okey so let's keep it.

danielpacak:
> Thank you for your review @danielpacak! 🎉 ❤️
> 
> > Before we merge this one, I'm wondering whether we should add a template to define the VulnerabilityReports custom resource. Some people say that the operator itself should not install the CRDs that it manages, but here we have a Helm chart, which provides means of installation.
> 
> Helm 3 supports this, but it is an evolving best practice. I think its the right call to bundle the CRDs with the helm chart. They won't be templates that render with values, and they won't be managed by Helm after install in general I think. For more info, see: https://helm.sh/docs/chart_best_practices/custom_resource_definitions/
> 
> * [x]  CRDs added
> * [x]  container securityContext runAsUser/runAsGroup updated to match predefined user/group in Dockerfile
> * [x]  Note about OPERATOR_NAMESPACE being inferred added.
> * [x]  Decide if `rbac.create: true` is to remain or be hardcoded as true. I'm not confident about the matter myself, but I suggest to let it remain unless better understanding about why so many helm chart has this setup and its no longer relevant for this helm chart.

Thanks for adding the crds to the chart. I didn't know that it's supported by Helm 3! Regarding caveats, I think we're good. If someone wants to managed CRD upgrades we'd suggest installing with OLM / https://operatorhub.io/operator/starboard-operator anyway.

danielpacak (approved):
Once again great job @consideRatio ! I'm going to merge the PR. As mentioned in the conversation we can follow up with in dedicated PR for such tasks as automated integration tests run as part of our CI workflow.

Regarding the support for multiple Helm releases running in the same cluster, I think we cannot do much about that. This problem is addressed by Operator Lifecycle Manager, where you define an OperatorGroup to configure multi-tenancy support of the operator. For example, if target namespaces specified by two different OperatorGroup instances intersect, the OLM won't validate such configuration.

I'll update docs / installation guid with Helm chart in this PR https://github.com/aquasecurity/starboard/pull/201

consideRatio:
Wieee! Thank you for your review and encouragement @danielpacak :heart: :tada: :sunflower:

PR #205, opened 2020-10-14
An extensible security toolkit for finding risks in your Kubernetes .README.md

sonam2905:


danielpacak:
Thank you for the PR @sonam2905 , but we already have all that information under the Introduction section.

PR #220, opened 2020-10-19
README.md Typo updated

echobash:


README.md
```diff
@@ -33,7 +33,7 @@
 Starboard integrates security tools into the Kubernetes environment, so that users can find and view the risks that
 relate to different resources in a Kubernetes-native way. Starboard provides
 [custom security resources definitions][starboard-crds] and a [Go module][starboard-go-module] to work with a range
-of existing security tools, as well as a `kubectl`-compatible command-line tool and an Octant plug-in that make security
+of existing security tools as well as a `kubectl`-compatible command-line tool and an Octant plug-in that makes security
```
danielpacak, line 36: We're referring to command-line tools and the Octant plugin so I believe we should use make instead of makes here

PR #225, opened 2020-10-23
Allow configuring custom Docker image for each scanner

### LINKED ISSUE #181: [cli] Allow configuring custom Docker image for each scanner

krisctl:
Fixes #181.

All unit tests and integration test passes. I also added some refactoring to bring the common Config interface in config.go and have one function to provide the image reference values. Also, I did not see anyone assigned to this issue and that's why created a pull request for it.

### DIFF
--- a/pkg/find/vulnerabilities/trivy/scanner.go
+++ b/pkg/find/vulnerabilities/trivy/scanner.go
@@ -27,12 +27,8 @@ import (
 	"k8s.io/utils/pointer"
 )
 
-type Config interface {
-	GetTrivyImageRef() string
-}
-
 // NewScanner constructs a new vulnerability Scanner with the specified options and Kubernetes client Interface.
-func NewScanner(config Config, opts kube.ScannerOpts, clientset kubernetes.Interface) *Scanner {
+func NewScanner(config starboard.Config, opts kube.ScannerOpts, clientset kubernetes.Interface) *Scanner {
 	return &Scanner{
 		config:    config,
 		opts:      opts,
@@ -43,7 +39,7 @@ func NewScanner(config Config, opts kube.ScannerOpts, clientset kubernetes.Inter
 }
 
 type Scanner struct {
-	config    Config
+	config    starboard.Config
 	opts      kube.ScannerOpts
 	clientset kubernetes.Interface
 	pods      *pod.Manager
@@ -135,7 +131,7 @@ func (s *Scanner) PrepareScanJob(_ context.Context, workload kube.Object, spec c
 	initContainers := []core.Container{
 		{
 			Name:                     initContainerName,
-			Image:                    s.config.GetTrivyImageRef(),
+			Image:                    s.config.GetImageRef(starboard.TrivyImageRef),
 			ImagePullPolicy:          core.PullIfNotPresent,
 			TerminationMessagePolicy: core.TerminationMessageFallbackToLogsOnError,
 			Env: []core.EnvVar{
@@ -248,7 +244,7 @@ func (s *Scanner) PrepareScanJob(_ context.Context, workload kube.Object, spec c
 
 		scanJobContainers[i] = core.Container{
 			Name:                     c.Name,
-			Image:                    s.config.GetTrivyImageRef(),
+			Image:                    s.config.GetImageRef(starboard.TrivyImageRef),
 			ImagePullPolicy:          core.PullIfNotPresent,
 			TerminationMessagePolicy: core.TerminationMessageFallbackToLogsOnError,
 			Env:                      envs,
--- a/pkg/kubebench/scanner.go
+++ b/pkg/kubebench/scanner.go
@@ -33,7 +33,7 @@ const (
 )
 
 type Config interface {
-	GetKubeBenchImageRef() string
+	GetImageRef(string) string
 }
 
 type Scanner struct {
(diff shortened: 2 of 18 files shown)


pkg/find/vulnerabilities/trivy/scanner.go
```diff
@@ -27,12 +27,8 @@ import (
 	"k8s.io/utils/pointer"
 )
 
-type Config interface {
```
danielpacak, line 30: This interface is supposed to defined other configuration options for vulnerability scanner, which are not currently implemented. The imageRef is only one property that might be configured.

What's more, it allows us to implement the interface segregation principle (part of SOLID). When we inject very generic starboard.Config interface into each scanner, we're violating this rule. In other words, we don't want to inject configuration of Polaris into a vulnerability scanner.

pkg/kubebench/scanner.go
```diff
@@ -33,7 +33,7 @@ const (
 )
 
 type Config interface {
-	GetKubeBenchImageRef() string
```
danielpacak, line 36: Don't break this interface. See my comment about SOLID and interface segregation.

pkg/kubehunter/scanner.go
```diff
@@ -25,22 +25,22 @@ import (
...
 const (
-	kubeHunterVersion       = "0.3.1"
 	kubeHunterContainerName = "kube-hunter"
 )
 
-var (
-	kubeHunterContainerImage = fmt.Sprintf("aquasec/kube-hunter:%s", kubeHunterVersion)
-)
+// KubeHunterVersion version of kube-hunter being used
+var KubeHunterVersion string
```
danielpacak, line 32: We should take version from the imageRef and avoid using global state / package level variables.

pkg/kubehunter/scanner.go
```diff
@@ -25,22 +25,22 @@ import (
...
 
 type Scanner struct {
+	config    starboard.Config
 	opts      kube.ScannerOpts
 	clientset kubernetes.Interface
 	pods      *pod.Manager
 }
 
-func NewScanner(opts kube.ScannerOpts, clientset kubernetes.Interface) *Scanner {
+func NewScanner(config starboard.Config, opts kube.ScannerOpts, clientset kubernetes.Interface) *Scanner {
```
danielpacak, line 41: Inject only kube-hunter config instead of potentially broad interface.

pkg/kubehunter/scanner.go
```diff
@@ -84,7 +84,13 @@ func (s *Scanner) Scan(ctx context.Context) (report starboardv1alpha1.KubeHunter
 	}()
 
 	// 4. Parse the KubeHuberOutput from the logs Reader
-	report, err = OutputFrom(logsReader)
+	version, err := starboard.GetVersionFromImageRef(s.config.GetImageRef(starboard.KubeHunterImageRef))
+	if err != nil {
+		err = fmt.Errorf("getting version from image ref: %w", err)
```
danielpacak, line 89: I think we should fails fast instead of implementing some kind of fallback logic here. If you return error someone will have a chance to fix the value in the ConfigMap.

pkg/operator/trivy/scanner.go
```diff
@@ -17,12 +17,12 @@ import (
...
 type trivyScanner struct {
 	idGenerator ext.IDGenerator
-	config      trivy.Config
+	config      starboard.Config
 }
 
 // NewScanner constructs a new VulnerabilityScanner, which is using an official
 // Trivy container image to scan pod containers.
-func NewScanner(idGenerator ext.IDGenerator, config trivy.Config) scanner.VulnerabilityScanner {
+func NewScanner(idGenerator ext.IDGenerator, config starboard.Config) scanner.VulnerabilityScanner {
```
danielpacak, line 25: Use trivy.Config for Trivy scanner

pkg/polaris/converter_test.go
```diff
@@ -5,20 +5,24 @@ import (
...
+	"github.com/aquasecurity/starboard/pkg/starboard"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
 
 func TestConverter_Convert(t *testing.T) {
 	// FIXME Deterministic assert!
 	t.Skip("Fix me - the assert is not deterministic")
+	config := starboard.ConfigData{
+		starboard.PolarisImageRef: starboard.GetDefaultConfig()[starboard.PolarisImageRef],
```
danielpacak, line 17: For testing purpose, we could just hardcode the key for setting imageRef.

pkg/polaris/scanner.go
```diff
@@ -29,22 +29,19 @@ import (
...
 type Scanner struct {
+	config    starboard.Config
 	opts      kube.ScannerOpts
 	clientset kubernetes.Interface
 	pods      *pod.Manager
 	converter Converter
 }
 
-func NewScanner(opts kube.ScannerOpts, clientset kubernetes.Interface) *Scanner {
+func NewScanner(config starboard.Config, opts kube.ScannerOpts, clientset kubernetes.Interface) *Scanner {
```
danielpacak, line 42: As mentioned above and other places, Polaris should not be aware of kube-bench or kube-hunter config. Define an interface for configuring Polaris, and let starboard.ConfigData implement this interface.

pkg/starboard/config.go
```diff
@@ -208,6 +213,10 @@ type BuildInfo struct {
 	Date    string
 }
 
+type Config interface {
```
danielpacak, line 216: We don't want to have such generic config because we plan to have more configuration options. Each scanner should define it's set of configuration settings.

danielpacak (changes requested):
Thanks for this PR @krisctl I left a few comments, but in general we'd like to separate configuration of vulnerability scanners from polaris or kube-hunter. That said, each scanner should have its own set of config settings agnostic of the config storage (e.g. ConfigMap).

krisctl:
Thanks for the feedback, @danielpacak. I will work on addressing these comments.

danielpacak:
Closing as stale / #181 has been resolved and adjusted to the current configuration mechanism based on ConfigMap and secret objects.

PR #193, opened 2020-10-10
refactor: Refactor get vulnerabilities command to use client instead of kubectl

hypnoglow:
Fixes https://github.com/aquasecurity/starboard/issues/142

Issue #142, opened 2020-09-02
The get vulnerabilities subcommand does not respect global flags (e.g. --kubeconfig)


danielpacak:
**What steps did you take and what happened:**

The global flags to configure client-go are not respected by the `get vulnerabilities` subcommand:

```
$ starboard --kubeconfig /path/to/kube/config \
    get vulnerabilities pod/nginx-tn2wq \
    -o yaml
```

**What did you expect to happen:**

The `--kubeconfig` flag and other global flags to be passed to the get vulnerabilities subcommand.

**Anything else you would like to add:**

N/A

**Environment:**

- Starboard version (use `starboard version`): 0.2.6
- Kubernetes version (use `kubectl version`): Any
- OS (macOS 10.15, Windows 10, Ubuntu 19.10 etc): Any

danielpacak:
@krisctl Are you available to work on this one? If not please let us know if we should leave it as unassigned.

neha-viswanathan:
@danielpacak I tried this out, and it seems like the ordering of the `kubeconfig` flag is an issue with your example:
```
kubectl starboard --kubeconfig=/path/to/my-kubeconfig get vulnerabilities pod/pod-xyz -o yaml                                      
apiVersion: v1
items: []
kind: List
metadata:
  resourceVersion: ""
  selfLink: ""
```

But the below one seems to work with passing the `kubeconfig` flag after the vulnerabilities subcommand.
```
kubectl starboard get vulnerabilities --kubeconfig=/path/to/my-kubeconfig pod/pod-xyz -o yaml
apiVersion: v1
items:
- apiVersion: aquasecurity.github.io/v1alpha1
  kind: VulnerabilityReport
  metadata:
    creationTimestamp: "2020-10-06T23:04:00Z"
    generation: 1
    labels:
-----
```

neha-viswanathan:
@danielpacak It seems like the example you have shared is missing the `namespace` flag. I tried to run the same command by providing the flag and I seem to get some output. Can you please share logs of what output/error you get when you run the command?

```
./bin/starboard --kubeconfig=/path/to/kubeconfig get vulnerabilities pod/pod-xyz -n=ns1 -o yaml
```

neha-viswanathan:
@danielpacak can you please share the logs from when you run the command?

danielpacak:
We don't have meaningful logs output as this command fails immediately due to connection error with Kubernetes API. To reproduce you can do the following:

```
[...]

PR #228, opened 2020-10-25
refactor: Refactor get configaudit command to use client instead of kubectl

hypnoglow:
The purpose is to solve the problem described in #142 but for configaudit.

Technically, this is similar to #193


danielpacak (approved):
LGTM! Thanks @hypnoglow !

PR #238, opened 2020-11-10
Starboard cli envars

immanetize:
Hey starboard team, I'm testing out starboard in an environment with enforced naming conventions.  This patch allows one to override the namespace and name of the starboard cli configmap using environment variables at runtime.

```
$ starboard config
error: configmaps "starboard" not found
$ STARBOARD_CONFIG_NAMESPACE=some-namespace starboard config |head
trivy.imageRef
docker.io/aquasec/trivy:0.9.1
trivy.severity
UNKNOWN,LOW,MEDIUM,HIGH,CRITICAL
kube-bench.imageRef
docker.io/aquasec/kube-bench:0.4.0
polaris.config.yaml
checks:
  # reliability
  multipleReplicasForDeployment: ignore
```


danielpacak:
> Hey starboard team, I'm testing out starboard in an environment with enforced naming conventions. This patch allows one to override the namespace and name of the starboard cli configmap using environment variables at runtime.
> 
> ```
> $ starboard config
> error: configmaps "starboard" not found
> $ STARBOARD_CONFIG_NAMESPACE=some-namespace starboard config |head
> trivy.imageRef
> docker.io/aquasec/trivy:0.9.1
> trivy.severity
> UNKNOWN,LOW,MEDIUM,HIGH,CRITICAL
> kube-bench.imageRef
> docker.io/aquasec/kube-bench:0.4.0
> polaris.config.yaml
> checks:
>   # reliability
>   multipleReplicasForDeployment: ignore
> ```

:wave: @immanetize Thank you for the PR and trying out Starboard in your environment!

Before we can merge it I'd like to better understand what do you mean by "enforcing naming conventions". Especially in the context of naming the ConfigMap created by the Starboard CLI on init. The reason I'm asking is that we wanted to keep this ConfigMap an **implementation detail** of the underlying Starboard config mechanism used by both CLI and the Operator.

Beyond that, the limitation of the current ConfigMap is that it should not be used for sensitive config values. However, we need to configure container image registry credentials or access tokens for Trivy run in client-server mode.

That said, I'm quite okey with making the starboard namespace configurable, but I'd think twice before we make the ConfigMap name configurable. As I said, it's likely to be replaced with a Secret or we'll keep both ConfigMap and Secret.

What's more, eventually we may have a CRD which represents Starboard's config and then based on such CR instance we'll create the ConfigMap and/or Secret indirectly. The CRD will allow us to validate the configuration of Starboard in declarative (OpenAPI Spec) or imperative way (validating admission controller). /cc @danielsagi @lizrice

danielpacak:
TL;DR; Namespace and configuration objects are implementation details of Starboard. Therefore I think we should not make them configurable. If it's still relevant, please start the discussion so we can collect more feedback from other users.

immanetize:
Hey Daniel, apologies for not seeing your replies sooner.  Your objections to these changes are unclear; I would think a future CRD implementation would have starboard look for it's configmap in whatever namespace it happens to be running in at the time, rather than hard code it...

To be a little more specific, I operate kubernetes clusters on which my security team would like to use Starboard.  We have [operator/crd] tooling to enable least privileged access to the cluster by provisioning rbac-enabled namespaces for each project - for example, members of the security team LDAP group would have administrative permissions on namespaces 'starboard-dev', 'starboard-staging', and 'starboard-production' namespaces.  Adjusting this permissions model and tooling seems relatively non-trivial... I urge you to reconsider, it's likely I'm not the only one where this design choice may preclude using starboard.  As I understand it, there's no current usage which would be broken by allowing deployment to varied namespaces.

danielpacak:
Thanks for providing more details @immanetize 

Currently the **Starboard Operator** is implemented in such way that it looks for the `starboard` ConfigMap and the `starboard` secret in the operator's namespace (i.e. the namespace in which the operator is installed and set via the `OPERATOR_NAMESPACE` env).

However it's not the case for **Starboard CLI**, which looks for config object in the hardcoded `starboard` namespace.

Admittedly, the configuration of **Starboard CLI** is limited, but intentionally we assumed that it would be enough for most users. For more advanced use cases we assumed that users will switch over to the **Starboard Operator** mode and stop using the CLI.

If I understood correctly, in your case you want to use Starboard CLI installed / initialized in different namespaces (e.g. `starboard-dev`, `starboard-qa`, etc.) and therefore the current implementation is not suitable for that.

Could you confirm that you may use both CLI and the operator?

immanetize:
I do recall some mixed usage, but that may have been part of choosing between the two patterns during POC and not strictly part of our use case.  I'll check in with the security folks using it and report back.

immanetize:
Hi, I can confirm mixed use.  We're interested in all of the reports that Starboard can produce, the operator only gives us vulnerability reports.

danielpacak:
Closing this one as there were many changes since this PR was created. Especially the way we configure plugins. Also the operator now generates ConfigAuditReports and CISKubeBenchReports. Please lets discuss first if the current configuration is still not suitable in your case.

Issue #261, opened 2020-12-04
Basic authentication support in Starboard Operator


lizrice:
Starboard CLI supports basic authentication for accessing images in a private registry. We should support this for the Operator as well

Issue #234, opened 2020-10-28
[operator] Cannot save VulnerabilityReport for pods owned by Node


danielpacak:

consideRatio:
For reference, this issue has a related discussion thread at https://github.com/aquasecurity/starboard/discussions/215#discussion-25235.

PR #267, opened 2020-12-06
feat(operator): Add basic authentication support for accessing images in private registries

danielpacak:
Resolves: #261


pkg/kube/runnable_job.go
```diff
@@ -66,30 +61,23 @@ func (r *runnableJob) Run(ctx context.Context) error {
...
+		secret.Namespace = r.job.Namespace
+		klog.V(3).Infof("Setting owner reference for secret %q", r.job.Namespace+"/"+secret.Name)
+		err = controllerutil.SetOwnerReference(r.job, secret, r.scheme)
 		if err != nil {
 			return err
 		}
-		klog.V(3).Infof("Updating secret %q", r.job.Namespace+"/"+s.Name)
-		r.secrets[i], err = r.clientset.CoreV1().Secrets(r.job.Namespace).Update(ctx, s, metav1.UpdateOptions{})
+		klog.V(3).Infof("Creating secret %q", r.job.Namespace+"/"+secret.Name)
+		r.secrets[i], err = r.clientset.CoreV1().Secrets(r.job.Namespace).Create(ctx, secret, metav1.CreateOptions{})
```
knqyf263, line 78: For my understanding: why does it work even though `Secret` is created after `Job` starts? IIUC, `Job` loads those secrets in the spec as environmental variables. I thought they should exist before Job starts.

danielpacak: Very good question. I think it works because a Job does not required secrets to be present when it's created. Only the pod that it controls requires those secrets to be present. Since it takes some time for Job controller to pick it up and create the pod, by that time the secrets are also created.

So I tested it and apparently works, but now I'm wondering if it might lead to some errors when a pod is scheduled very very fast.

knqyf263: Yes, I'm worried about that. It may also take longer for some other reason to create `Secret`. `Job` might create the Pod before creating `Secret`. It works in many cases as you said, but it might cause an unclear error occasionally.

danielpacak: Okey let me revert that for the CLI and do the same for the operator:

- Create Secrets referred to by the Job
- Create the Job
- Update Secrets' ownerReferences to point to the Job

This is an extra Kubernetes API cal but more readable and predictable.

knqyf263: I'm not sure it is possible to happen. If you can say it never happens, you don't have to change the order. I'm just concerned.

knqyf263 (approved):

pkg/kube/runnable_job.go
danielpacak: I think it may happen even though it's not very likely. Anyway I've changed that to order to first create secrets and then job.

knqyf263: I've already approved :shipit:

shanoor:
Hi, thanks for this, anyone knows when the version including this fix will release?

danielpacak:
> Hi, thanks for this, anyone knows when the version including this fix will release?

We don't have a strict release schedule, but I think this feature is useful so I plan it for [Release v0.8.0](https://github.com/aquasecurity/starboard/milestone/7) There's no planned release date, but I think we can ship it once #234 is resolved.

PR #271, opened 2020-12-08
refactor: Scanner CLI can easily swap the implementations of the underlying scanner

danielpacak:

### DIFF
--- a/pkg/find/vulnerabilities/scanner.go
+++ b/pkg/find/vulnerabilities/scanner.go
@@ -14,7 +14,6 @@ import (
 	"github.com/aquasecurity/starboard/pkg/runner"
 	"github.com/aquasecurity/starboard/pkg/scanners"
 	"github.com/aquasecurity/starboard/pkg/starboard"
-	"github.com/aquasecurity/starboard/pkg/trivy"
 	"github.com/aquasecurity/starboard/pkg/vulnerabilityreport"
 	batchv1 "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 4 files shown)


pkg/trivy/scanner.go
```diff
@@ -315,10 +315,11 @@ func (s *scanner) getPodSpecForClientServerMode(spec corev1.PodSpec, credentials
...
 
 		containers = append(containers, corev1.Container{
-			Name:            container.Name,
-			Image:           s.config.GetTrivyImageRef(),
-			ImagePullPolicy: corev1.PullIfNotPresent,
-			Env:             env,
+			Name:                     container.Name,
+			Image:                    s.config.GetTrivyImageRef(),
+			ImagePullPolicy:          corev1.PullIfNotPresent,
+			TerminationMessagePolicy: corev1.TerminationMessageFallbackToLogsOnError,
```
danielpacak, line 321: We could not see the error logs coming from pod logs in case of failure. (In ClientServer mode)

pkg/find/vulnerabilities/scanner.go
```diff
@@ -25,39 +24,45 @@ import (
...
+	ext.IDGenerator
+	vulnerabilityreport.Plugin
 	kube.SecretsReader
 }
 
-// NewScanner constructs a new vulnerability Scanner with the specified options and Kubernetes client Interface.
+// NewScanner constructs a new static vulnerability Scanner with the specified
+// vulnerabilityreport.Plugin that knows how to perform the actual scanning,
+// which is performed by running a Kubernetes job, and knows how to convert logs
+// to instances of v1alpha1.VulnerabilityReport.
```
lizrice, line 42: I wonder if this eventually becomes an even more generic Scanner that is configured to run a particular type of scan, and convert to a configurable type of report? Doesn't need to happen now in any case, I'm just wondering about it

lizrice (approved):
lgtm

pkg/find/vulnerabilities/scanner.go
danielpacak: All scanners look similar in terms of creating a job, waiting for its completion, parsing logs, and finally writing report to etcd. So yes, maybe one day we can generalize this code as you said.

PR #276, opened 2020-12-14
feat(config): Add starboard secret to store sensitive config params

danielpacak:
The init command creates the starboard ConfigMap in the starboard namespace.
This is not suitable for setting sensitive params such as tokens or passwords.

### DIFF
--- a/pkg/starboard/config.go
+++ b/pkg/starboard/config.go
@@ -3,6 +3,7 @@ package starboard
 import (
 	"context"
 	"fmt"
+	"k8s.io/apimachinery/pkg/labels"
 
 	starboardv1alpha1 "github.com/aquasecurity/starboard/pkg/apis/aquasecurity/v1alpha1"
 	appsv1 "k8s.io/api/apps/v1"
@@ -12,7 +13,6 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/client-go/kubernetes"
 
 	"github.com/google/go-containerregistry/pkg/name"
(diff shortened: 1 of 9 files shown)


pkg/starboard/config.go
```diff
@@ -332,7 +332,21 @@ func (c *configManager) EnsureDefault(ctx context.Context) error {
...
+
+	secret := &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace: c.namespace,
+			Name:      SecretName,
+			Labels: labels.Set{
+				"app.kubernetes.io/managed-by": "starboard",
+			},
+		},
+	}
```
knqyf263, line 347: It looks like the secret is initialized with empty data. Where are values filled? A user edits it manually?

danielpacak: Yup. It is supposed to be UPDATED or PATCHED manually with kubectl (Starboard CLI) or programmatically with Kubernetes API (Aqua Wave)

knqyf263: For example, a user needs to set `trivy.githubToken`, right?
https://github.com/aquasecurity/starboard/blob/bbd497b0df6755c6c05cd286bbafd89e93a24ff6/pkg/trivy/scanner.go#L128

Does Starboard CLI support to set it? Or, does a user configure it via raw kubectl?

danielpacak: The starboard.ConfigManager is used by Starboard CLI and the operator. And yes, users are supposed to edit the secret with `kubectl` until we implement `starboard config set -from-literal=key1=config1`. Currently we just have a command to read config and hide implementation details `starboard config get --name-only`.

The configuration steps are documented in https://github.com/aquasecurity/starboard#configuration
I will update it in another PR while migrating from README to MkDocs, but the only difference is that we introduced secret to configure sensitive params such as `trivy.githubToken`.

Beyond that, there will be more values that we should not store in ConfigMap such as Trivy server tokens, Aqua CSP credentials, etc.

knqyf263: I didn't know the document.
https://github.com/aquasecurity/starboard#configuration

It makes sense. Thanks.

knqyf263 (approved):

PR #286, opened 2020-12-18
refactor: Unify scanners config

danielpacak:
- Starboard CLI & Starboard Operator read config from ConfigMap and secret
- Starboard CLI supports Aqua scanner!
- Documentation


docs/configuration.md
```diff
@@ -16,44 +16,64 @@ For example, by default Trivy displays vulnerabilities with all severity levels
...
-| `trivy.severity`          | `UNKNOWN,LOW,MEDIUM,HIGH,CRITICAL`                     | A comma separated list of severity levels reported by Trivy |
-| `trivy.imageRef`          | `docker.io/aquasec/trivy:0.14.0`                       | Trivy image reference |
-| `trivy.mode`              | `Standalone`                                           | Trivy client mode. Either `Standalone` or `ClientServer`. Depending on the active mode other settings might be applicable or required. |
-| `trivy.serverURL`         | `http://trivy-server.trivy-server:4954`                | The endpoint URL of the Trivy server. Required in `ClientServer` mode. |
-| `trivy.serverTokenHeader` | `Trivy-Token`                                          | The name of the HTTP header to send the authentication token to Trivy server. Only application in `ClientServer` mode when `trivy.serverToken` is specified. |
-| `polaris.config.yaml`     | [Check the default value here][default-polaris-config] | Polaris configuration file |
+| CONFIGMAP KEY                  | DEFAULT                                                | DESCRIPTION |
```
lizrice, line 52: I wonder if we should explicitly say somewhere that you only need to configure the settings for the scanner you're using (i.e. trivy.* is only used if scanner is set to Trivy)

lizrice: Maybe also a reference to the Integrations section where it's all laid out clearly what you need to do with these settings

docs/integrations.md
```diff
@@ -1,5 +1,139 @@
...
+## Vulnerability Scanners
+
+![](images/integration-vulnerability-scanners.png)
+
+### Trivy Standalone
+
+The default configuration settings enable Trivy `vulnerabilityReports.scanner` in [`Standalone`][trivy-standalone]
+`trivy.mode`. Even though it doesn't require any additional setup, it's the least efficient method. Each pod created
+by a scan job has the init container that downloads the Trivy vulnerabilities database from the GitHub releases page
+and stores it in the local file system of an [emptyDir][emptyDir-volume] volume. Such volume is then shared with main
```
lizrice, line 12:
```suggestion
and stores it in the local file system of an [emptyDir][emptyDir-volume] volume. This volume is then shared with the main
```

docs/integrations.md
```diff
@@ -1,5 +1,139 @@
...
+containers that perform the actual scanning. Finally, the pod is deleted along with the emptyDir volume.
+
+![](images/design/trivy-standalone.png)
+
+The number of containers defined by a scan job equals the number of containers defined by the scanned Kubernetes
+workload, so the cache in this mode is useful only if the workload defines multiple containers.
+
+Beyond that, frequent downloads from GitHub might lead to [rate limiting][gh-rate-limiting] problem. The limits are
```
lizrice, line 20:
```suggestion
Beyond that, frequent downloads from GitHub might lead to a [rate limiting][gh-rate-limiting] problem. The limits are
```

docs/integrations.md
```diff
@@ -1,5 +1,139 @@
...
+```
+
+[trivy-standalone]: https://github.com/aquasecurity/trivy#standalone
+[emptyDir-volume]: https://kubernetes.io/docs/concepts/storage/volumes/#emptydir
+[gh-rate-limiting]: https://docs.github.com/en/free-pro-team@latest/rest/overview/resources-in-the-rest-api#rate-limiting
+
+### Trivy ClientServer
+
+You can connect Starboard to an external [Trivy server][trivy-clientserver] by changing the default `trivy.mode` from
+[`Standalone`][trivy-standalone] to [`ClientServer`][trivy-clientserver] and specifying `trivy.serverURL`.
```
lizrice, line 46: Maybe worth adding something like 
```
The Trivy Server could be your own deployment, or it could be an external service. See <Trivy docs> for more information on deploying Trivy in Client-Server mode.
```

pkg/starboard/config.go
```diff
@@ -280,6 +327,15 @@ func (c ConfigData) GetKubeBenchImageRef() string {
 	return "docker.io/aquasec/kube-bench:0.4.0"
```
lizrice, line 327: could we get this from the default config settings? Would be nice to only have the default images defined in one place

lizrice (approved):
I made a couple of suggestions but otherwise lgtm!

docs/configuration.md
danielpacak: Good point. I added both.

docs/integrations.md
danielpacak: Done

pkg/starboard/config.go
danielpacak: Certainly. I fixed it so the default is set once. When we read, but the property is not set we return error.

Issue #244, opened 2020-11-23
Jobs try to run on Windows


sylus:
**What steps did you take and what happened:**

I am running a cronjob which calls starboard with -v3 kube-bench but notice that some jobs are unable to run as they are trying to execute on windows nodes.

P.S. Thanks for the amazing tool :D 

**What did you expect to happen:**

The windows nodes are tainted.

**Anything else you would like to add:**

N/A

**Environment:**

- Starboard version (use `starboard version`): 0.6.0
- Kubernetes version (use `kubectl version`): 1.17.3
- OS (macOS 10.15, Windows 10, Ubuntu 19.10 etc): macOS 10.15

danielpacak:
👋 @sylus Thank you for opening this issue. What would be desired behaviour in your case? Ignoring windows nodes (or in general any tainted node) and run kube-bench only on linux nodes?

sylus:
Yup @danielpacak i think that would be the solution i'd like to see :D

krol3:
I'd like to test this, and try to do the fixes for this issue. 
Thanks, @sylus, It seems that kube-bench doesn’t  support windows nodes. [More details](https://github.com/aquasecurity/kube-bench/issues/624)

PR #289, opened 2020-12-23
Issue 244 - Add affinity to the linux node by default

krol3:
This PR add a default affinity to create the pod in the linux nodes.

Related the [issue-244](https://github.com/aquasecurity/starboard/issues/244)

### DIFF
--- a/pkg/cmd/scan_ciskubebench.go
+++ b/pkg/cmd/scan_ciskubebench.go
@@ -70,23 +70,27 @@ func ScanKubeBenchReports(cf *genericclioptions.ConfigFlags) func(cmd *cobra.Com
 		var wg sync.WaitGroup
 
 		for _, node := range nodeList.Items {
+
+			nodeValueLabel, exist := node.GetObjectMeta().GetLabels()["kubernetes.io/os"]
+			if exist && nodeValueLabel != "linux" {
+				klog.V(3).Infof("Skipping non linux node: %v %v", node.Name, node.Labels)
+				continue
+			}
+
 			wg.Add(1)
 			go func(node corev1.Node) {
 				defer wg.Done()
 
-				nodeValueLabel, exist := node.GetObjectMeta().GetLabels()["kubernetes.io/os"]
-				if exist && nodeValueLabel == "linux" {
-					report, err := scanner.Scan(ctx, node)
+				report, err := scanner.Scan(ctx, node)
 
-					if err != nil {
-						klog.Errorf("Error while running kube-bench on node: %s: %v", node.Name, err)
-						return
-					}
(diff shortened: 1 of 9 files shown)


pkg/cmd/scan_ciskubebench.go
```diff
@@ -74,16 +74,19 @@ func ScanKubeBenchReports(cf *genericclioptions.ConfigFlags) func(cmd *cobra.Com
 			go func(node corev1.Node) {
 				defer wg.Done()
 
-				report, err := scanner.Scan(ctx, node)
+				nodeValueLabel, exist := node.GetObjectMeta().GetLabels()["kubernetes.io/os"]
+				if exist && nodeValueLabel == "linux" {
+					report, err := scanner.Scan(ctx, node)
```
danielpacak, line 79: I think it's even better to avoid creating a gorutine when we detect that's it non linux node. Also we could log warning when we skip Windows nodes:

```go
for _, node := range nodeList.Items {

	if !isLinuxNode(node) {
		klog.Warning("Skipping non linux node: %v %v", node.Name, node.Labels)
		continue
	}

	wg.Add(1)

	// ...
}
```

I prefer exiting the loop asap (using `continue` keyword) to embedding too many `if` statements and shifting the code to the right.

pkg/kubebench/scanner.go
```diff
@@ -149,6 +149,7 @@ func (s *Scanner) prepareKubeBenchJob(node corev1.Node) (*batchv1.Job, error) {
 					RestartPolicy:                corev1.RestartPolicyNever,
 					HostPID:                      true,
 					NodeName:                     node.Name,
+					Affinity:                     starboard.DefaultAffinity(),
```
danielpacak, line 152: For kube-bench we explicitly set the NodeName property. If we avoid creating a scan job for windows nodes the node affinity would be redundant.

pkg/starboard/config.go
```diff
@@ -431,3 +431,23 @@ func (c *configManager) Delete(ctx context.Context) error {
 	}
 	return nil
 }
+
+// DefaultAffinity constructs a new Affinaty resource with default values: linux supported nodes
+func DefaultAffinity() *corev1.Affinity {
```
danielpacak, line 436: I think the intention is to explicitly schedule to linux nodes. The default is not very meaningful until you read the code of the method.

```suggestion
func LinuxNodeAffinity() *corev1.Affinity {
```

danielpacak (changes requested):
@krol3 I left a few comments. I also notices that there are failing unit tests.

danielpacak (approved):

PR #317, opened 2021-01-08
add integration tests target

### LINKED ISSUE #316: Contributing guidelines refer to non-existent `integration-tests` make target

`CONTRIBUTING.md` says:

> To run all integration tests with code coverage enabled, run:
> ```
> $ make integration-tests
> ```

Running this command in the shell, I get:

```
$ make integration-tests
make: *** No rule to make target 'integration-tests'.  Stop.
```

I was expecting the integration tests to run.

**Environment:**

I ran the command on a fresh clone of the repository, in the root of the repository, on the `main` branch.

mlevesquedion:
Fixes #316.

I'm assuming that `CONTRIBUTING.md` represents the desired state, that is to say an `integration-tests` target is desired. If that's not the case, I'm happy to edit `CONTRIBUTING.md` to indicate how to actually run the integration tests.


danielpacak (changes requested):
:wave: @mlevesquedion Thanks for spotting this out and opening PR! I think I'd rather update CONTRIBUTING.md and explain that we have separate test suits for Starboard CLI and Starboard Operator.

We run them separately in our CI workflow after all:
- https://github.com/aquasecurity/starboard/blob/main/.github/workflows/build.yaml#L68
- https://github.com/aquasecurity/starboard/blob/main/.github/workflows/build.yaml#L97

Beyond that, the first suite may leave the cluster in dirty state, and we may end up with flaky integration tests. WDYT?

mlevesquedion:
Hello @danielpacak!

That makes sense to me. Thanks for pointing out the workflows.

I've removed the `make` target and updated `CONTRIBUTING.md` instead. PTAL.

danielpacak (approved):
Looks great! Thank you @mlevesquedion !

PR #318, opened 2021-01-08
docs(CONTRIBUTING.md): Update OLM guides

danielpacak:
- The main change is how we build catalog of operators to test the operator before it's published to the community operators repo. Previously we had to publish to Quay.io apps in some proprietary format. Currently it's based on CatalogSource resource and Docker image. Nice and simple.
- I also deleted the `./deploy/olm/bundle` directory which is redundant - we already have the master copy at https://github.com/operator-framework/community-operators/tree/master/upstream-community-operators/starboard-operator


CONTRIBUTING.md
```diff
@@ -207,174 +208,185 @@ started with a basic development workflow. For other install modes see [Operator
...
+NAME                 CATALOG             AGE
+starboard-operator   Starboard Catalog   97s
+```
 
-   An OperatorSource resource defines the external data store used to host operator bundles. In this case, you will be
-   defining an OperatorSource to point to your Quay.io account, which will provide access to its hosted OLM bundles.
+If the Starboard Operator appears in this list, the catalog was successfully parsed and it is now available to install.
 
-2. Create the OperatorGroup resource:
+### Create the OperatorGroup
```
lizrice, line 305: Is some of this duplicate of what's at https://aquasecurity.github.io/starboard/operator/installation/olm/ ? I guess here in the contribution guide it makes sense to talk about how to build and publish it but if from here on it's about running it, maybe refer to the documentation for that?

lizrice (approved):
I think one section maybe duplicates installation info that's also in the main docs? otherwise fine

CONTRIBUTING.md
danielpacak: Yup. It's duplicate. The only difference is the source catalog `starboard-catalog` instead of `operatorhubio-catalog` in the Subscription config (`spec.source`). I removed duplicate and added link to OLM installation guide.

PR #319, opened 2021-01-08
docs(olm): Update docs how to install operator from OperatorHub.io

danielpacak:
in the [lates version](https://operatorhub.io/operator/starboard-operator) we changed the way vulnerability scanners are configured.  Instead of using env the operator reads config objects in the namespace where it's installed.

You can preview docs by running `make mkdocs-serve`

### DIFF
--- a/docs/cli/installation/docker.md
+++ b/docs/cli/installation/docker.md
@@ -1,7 +1,7 @@
-We also release the Docker image `aquasec/starboard:$VERSION` to run Starboard as a Docker container or to
-manually schedule Kubernetes scan Jobs in your cluster.
+We also release Docker images `aquasec/starboard:$VERSION` and `public.ecr.aws/aquasecurity/starboard:$VERSION` to run
+Starboard as a Docker container or to manually schedule Kubernetes scan Jobs in your cluster.
 
 ```
-docker container run --rm aquasec/starboard:0.4.0 version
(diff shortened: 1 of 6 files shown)


lizrice (commented):
One small thing - when you expand the Starboard Operator menu its last item is "Configuration" (of the operator) but then the next top-level item is also "Configuration", which looks a bit confusing! Wonder if it makes sense to rename the top level item - maybe something like Starboard Settings? 

And then maybe the first sentence of that could be something like:

> The Starboard CLI and Starboard Operator both read their configuration settings from a ConfigMap, as well as a Secret that holds confidential settings (such as a GitHub token). The `starboard init` command....

Screenshot 2021-01-08 at 19 36 59

danielpacak:
> One small thing - when you expand the Starboard Operator menu its last item is "Configuration" (of the operator) but then the next top-level item is also "Configuration", which looks a bit confusing!
> 
> Screenshot 2021-01-08 at 19 36 59
> 
> Wonder if it makes sense to rename the top level item - maybe something like Starboard Settings?
> 
> And then maybe the first sentence of that could be something like:
> 
> > The Starboard CLI and Starboard Operator both read their configuration settings from a ConfigMap, as well as a Secret that holds confidential settings (such as a GitHub token). The `starboard init` command....

I agree that there's too many Configurations :) Renamed to **Starboard Settings**

lizrice (approved):

PR #331, opened 2021-01-12
Add kubeHunter.quick configuration parameter

### LINKED ISSUE #165: Figure out why integration test for kube-hunter command is failing on GitHub actions runner
We have an integration test which runs kube-hunter command. However this tests fails when it's run on GitHub actions runner. It seems to be working fine in dev environment with local KIND cluster though.

mlevesquedion:
Fixes #165.

This PR adds a `kubeHunter.quick` key to starboard's configuration settings. When the `quick` key is set to `"true"`, `kube-hunter` is called with the `--quick` flag, which enables quick subnet scanning (/24). By default, `quick` scanning is disabled.

Setting this key to `"true"` in the integration test for `starboard` allows the test for `kube-hunter` to run in a reasonable amount of time. Indeed, without the `--quick` flag, `kube-hunter` will attempt to scan an entire `/16` CIDR, which takes a lot of time.

I used `"false"` and `"true"` (the strings `"false"` and `"true"`, not the booleans `false` and `true`) as possible values for the `kubeHunter.quick` config key, because currently the `ConfigData` type only allows `string` values:

```go
// ConfigData holds Starboard configuration settings as a set
// of key-value pairs.
type ConfigData map[string]string
```

However, YAML has booleans, so this may seem a little weird to users. Indeed, attempting to use `true` instead of `"true"` will cause a weird error: `Invalid value: "map[...": unrecognized type: string`.

One approach to circumvent this issue would be to use the strings "no" and "yes" instead.

Let me know what you think.


danielpacak (approved):
LGTM! Thanks a lot @mlevesquedion for resolving the mystery of kube-hunter run on GitHub / Azure and adding this config 🚀

PR #332, opened 2021-01-12
docs(crds): Overview custom resource definitions used by Starboard

danielpacak:
To start with it's an overview. Later on we can add examples and explain the contents of each report.

Follow up on https://github.com/aquasecurity/starboard/discussions/324


docs/crds.md
```diff
@@ -1,18 +1,69 @@
...
+
+Each report owned by the underlying Kubernetes workload and is stored in the same namespace, following the
+`<workload-kind>-<workload-name>` naming convention.
+
+Third party Kubernetes configuration checkers, linters, and sanitizers that are compliant with the ConfigAuditReport
+schema can be integrated with Starboard.
+
+> **NOTE** The challenge with onboarding third party configuration checkers is that they tend to have different interfaces
+> to perform scans and vary in output formats for a relatively common goal, which is inspecting deployment descriptors
+> for common configuration pitfalls.
```
lizrice, line 41: I think we should add something here about being open to suggestions for changes to the CRD in the case that would enable additional third-party tools

lizrice (approved):
One request but LGTM

docs/crds.md
danielpacak: I added similar sentence as a note in the top of this page. I think we open for changes to all CRDs, not only ConfigAdutiReport

PR #333, opened 2021-01-12
refactor: Define interface for resource configuration checkers

danielpacak:
This is a preliminary step to onboard Popeye or any other configuration auditing tool and reuse code to integrate Polaris with Starboard Operator.

### DIFF
--- a/pkg/configauditreport/plugin.go
+++ b/pkg/configauditreport/plugin.go
@@ -1,0 +1,21 @@
+package configauditreport
+
+import (
+	"io"
+
+	"github.com/aquasecurity/starboard/pkg/apis/aquasecurity/v1alpha1"
+	"github.com/aquasecurity/starboard/pkg/kube"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+// Plugin defines the interface between Starboard and Kubernetes workload
+// configuration checkers / linters / sanitizers. Not a final version, rather
+// first step to separate generic workloads discovery code and Polars
+// implementation details.
+type Plugin interface {
+	GetScanJobSpec(workload kube.Object, gvk schema.GroupVersionKind) (corev1.PodSpec, error)
+
+	ParseConfigAuditResult(logsReader io.ReadCloser) (
+		v1alpha1.ConfigAuditResult, error)
+}
--- a/pkg/configauditreport/plugin.go
+++ b/pkg/configauditreport/plugin.go
@@ -11,7 +11,7 @@ import (
 
 // Plugin defines the interface between Starboard and Kubernetes workload
 // configuration checkers / linters / sanitizers. Not a final version, rather
-// first step to separate generic workloads discovery code and Polars
+// first step to separate generic workloads discovery code and Polaris
 // implementation details.
 type Plugin interface {
 	GetScanJobSpec(workload kube.Object, gvk schema.GroupVersionKind) (corev1.PodSpec, error)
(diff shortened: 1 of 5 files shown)


pkg/polaris/scanner.go
```diff
@@ -28,32 +29,29 @@ const (
 	configVolume         = "config"
 )
 
-type Config interface {
-	GetPolarisImageRef() (string, error)
-}
-
+// TODO Move the Scanner struct to configauditreport package. It's generic and can be used with similar tools to Polaris.
```
danielpacak, line 32: I will do that in a separate PR. Otherwise reviewing the changes would be difficult.

pkg/vulnerabilityreport/scanner.go
```diff
@@ -29,9 +29,9 @@ type Scanner struct {
 	scheme    *runtime.Scheme
 	clientset kubernetes.Interface
 	opts      kube.ScannerOpts
+	plugin    Plugin
 	pods      *pod.Manager
 	ext.IDGenerator
-	Plugin
```
simar7, line 34: Any reason why it's better to name it? Purely subjective, but I do think implicit encapsulation helps reduce boilerplate like `s.plugin.XXX`. Thoughts?

pkg/polaris/scanner_test.go
```diff
@@ -0,0 +1,103 @@
...
+			},
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			plugin := polaris.NewPlugin(tc.config)
+			jobSpec, err := plugin.GetScanJobSpec(tc.workload, tc.gvk)
+			require.NoError(t, err)
+			assert.Equal(t, tc.expectedJobSpec, jobSpec)
```
simar7, line 99: You could add `tc.name` as last param to both these calls as it would display the test name. I've found it helpful while debugging table driven tests.

pkg/configauditreport/plugin.go
```diff
@@ -0,0 +1,21 @@
...
+
+	"github.com/aquasecurity/starboard/pkg/apis/aquasecurity/v1alpha1"
+	"github.com/aquasecurity/starboard/pkg/kube"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+// Plugin defines the interface between Starboard and Kubernetes workload
+// configuration checkers / linters / sanitizers. Not a final version, rather
+// first step to separate generic workloads discovery code and Polars
```
simar7, line 14:
```suggestion
// first step to separate generic workloads discovery code and Polaris
```

pkg/configauditreport/plugin.go
```diff
@@ -0,0 +1,21 @@
...
+// implementation details.
+type Plugin interface {
+	GetScanJobSpec(workload kube.Object, gvk schema.GroupVersionKind) (corev1.PodSpec, error)
+
+	ParseConfigAuditResult(logsReader io.ReadCloser) (
+		v1alpha1.ConfigAuditResult, error)
```
simar7, line 20: Any reason not to single line it?

simar7 (approved):
Some minor nits but overall lgtm

pkg/vulnerabilityreport/scanner.go
danielpacak: I just want to quickly identify that this call / functionality is contributed by a plugin and distinguish it from the generic code. I also don't want to export the Plugin field.

danielpacak: But in general I prefer shorter syntax

pkg/polaris/scanner_test.go
danielpacak: Nice! Added

pkg/configauditreport/plugin.go
danielpacak: Nope. Fixed

PR #350, opened 2021-01-17
docs: Use admonitions

danielpacak:
admonition_tip

admonition_warning


docs/operator/troubleshooting.md
```diff
@@ -0,0 +1 @@
+TBD
```
lizrice, line 1: It might be better to leave this section out altogether (unless you're planning on adding something here imminently)

lizrice (approved):
Looks nice! I think it might be better to leave out the "TBD" troubleshooting guide for the Operator unless you know what's going into it

docs/operator/troubleshooting.md
danielpacak: Good point. I'm removing it for now and will add it back once we get some feedback from end users.

Issue #163, opened 2020-09-30
Set SecurityContext for pods created by scan jobs


danielpacak:
To implement least privileged principle we should explicitly set SecurityContexts<sup>[1]</sup> on pod templates for each scan job. Such SecurityContext should reflect minimal privileges to run a given scanner. The goal here is to review required permissions and set them explicitly for each pod template comprising a scan job descriptor.

[1]: https://kubernetes.io/docs/tasks/configure-pod-container/security-context/

akath19:
@danielpacak I'd like to pick this one up, here's my suggested SecurityContext:

Pod Level:
```
securityContext:
   runAsUser: 1900
   runAsGroup: 1900
   fsGroup: 1900
   runAsNonRoot: true
```

Container level:

```
securityContext:
  allowPrivilegeEscalation: false
  capabilities:
      drop:
       - all
  privileged: false
  readOnlyRootFilesystem: false
````

I've tested this with trivy & polaris jobs, I haven't tested this with kube-bench/hunter jobs yet.

danielpacak:
👋 @akath19 Happy to assign this one to you!

I think we can start with hardcoded structure literals to specify SecurityContexts for scan jobs. Later on we can think of making it configurable.

For the SecurityContext itself please go ahead and open a PR so I can review it. I believe we cannot run Trivy as non root user, but if that's the case our integration should catch it.

danielpacak:
@akath19 Any update or blocker on this one? Shall I keep it assigned to you?

akath19:
Sorry about that, I've been super busy at work.

I already have a basic PR for this, I'll send it soon

danielpacak:
> Sorry about that, I've been super busy at work.
> 
> I already have a basic PR for this, I'll send it soon

No worries and no rush. It's just a ping. Feel free to split this issue into smaller PRs, e.g. for each scanner. That way we can make it less overwhelming and simpler to review.

lizrice:
Hi @akath19, we're unassigning this for now so that someone else could pick it up. If your PR is ready or nearly ready and you can submit it soon please let us know and we'll assign it back to you. Thanks!

mlevesquedion:
I would be interested in picking this up.

danielpacak:
That's great @mlevesquedion . Please go ahead and consider splitting this task into smaller pull requests, e.g. review SecurityContext for each scanner one by one.

mlevesquedion:
[...]

PR #354, opened 2021-01-20
Set security context for kube-bench

mlevesquedion:
This PR adds `SecurityContext`s to the `kube-bench` `Job` in order to run the job with the least amount of privilege possible, in partial fulfillment of #163.

I ran the job with and without the changes in a local `kind` cluster and I did not see a difference in the output.

### DIFF
--- a/pkg/kubebench/scanner.go
+++ b/pkg/kubebench/scanner.go
@@ -134,6 +134,13 @@ func (s *Scanner) prepareKubeBenchJob(node corev1.Node) (*batchv1.Job, error) {
 					RestartPolicy:                corev1.RestartPolicyNever,
 					HostPID:                      true,
 					NodeName:                     node.Name,
+					SecurityContext: &corev1.PodSecurityContext{
+						RunAsUser:  pointer.Int64Ptr(0),
+						RunAsGroup: pointer.Int64Ptr(0),
+						SeccompProfile: &corev1.SeccompProfile{
+							Type: corev1.SeccompProfileTypeRuntimeDefault,
+						},
+					},
 					Volumes: []corev1.Volume{
 						{
 							Name: "var-lib-etcd",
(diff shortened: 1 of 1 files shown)


danielpacak:
👋 @lizrice Would you help reviewing this one? I want to make sure that by changing permissions we'd not show wrong results.

lizrice:
Just an update that I'm seeing some issues, very probably more related to [a kube-bench problem](https://github.com/aquasecurity/kube-bench/issues/768) than to this PR, but I'd like us to get to the bottom of that first before I can confirm this PR

lizrice:
Unfortunately this is going to cause a problem, because the non-root user won't have permission to look at the kubelet config file /var/lib/kubelet/config.yaml when it tries to run kube-bench on worker nodes. It's currently not doing that on single-node clusters because of #369 😢

I hacked together a local fix for #369 to check, and saw the following if using this security context: 

```
bin/starboard scan ciskubebenchreports -v 3
I0128 17:55:36.016807  109627 runner.go:79] Running task and waiting forever
I0128 17:55:36.016989  109627 runnable_job.go:76] Creating job "starboard/5a60aa83-278f-4715-ab58-2f998c9d4930"
I0128 17:55:36.064560  109627 runnable_job.go:132] Event: Created pod: 5a60aa83-278f-4715-ab58-2f998c9d4930-p67m2 (SuccessfulCreate)
I0128 17:55:42.071328  109627 runnable_job.go:114] Stopping runnable job on task failure with status: Failed
E0128 17:55:42.086030  109627 runnable_job.go:164] Container kube-bench terminated with Error:
error looking for file /var/lib/kubelet/config.yaml: stat /var/lib/kubelet/config.yaml: permission denied
```

mlevesquedion:
Got it, thank you for describing the issue! I'll see how I can update the `SecurityContext` to keep the job working while restricting privilege as much as possible.

danielpacak:
@mlevesquedion We've already fixed #369 so I think we can proceed with this PR. Notice there's a conflict to be resolved.

mlevesquedion:
I've resolved the conflict and I updated the security context to be able to read the `/var/lib/kubelet/config.yaml`. Unfortunately, in the general case, in order to be able to read that file, the container will have to run as root. I've made that explicit in the security context.

danielpacak (approved):
LGTM I tested with KIND cluster with 2 control plane and 3 worker nodes and everything is fine. Thank you @mlevesquedion for help!

PR #358, opened 2021-01-22
chore: Update deployment descriptors and docs

danielpacak:
Apologies for mixing changes to docs and other chores, but I was reading the docs to test the operator.


deploy/helm/values.yaml
```diff
@@ -50,12 +50,6 @@ envSecret:
   #       properly.
   stringData:
     OPERATOR_LOG_DEV_MODE: "false"
-    OPERATOR_SCANNER_TRIVY_ENABLED: "true"
```
danielpacak, line 53: Not used anymore. We store Aqua Enterprise settings in the `starboard` secret.

deploy/crd/configauditreports.crd.yaml
```diff
@@ -20,3 +20,23 @@ spec:
       - all
     shortNames:
       - configaudit
+  additionalPrinterColumns:
+    - JSONPath: .report.scanner.name
+      type: string
+      name: Scanner
```
lizrice, line 26:
```suggestion
      name: Scanner
      description: The name of the config audit scanner
```

To match what we have for vuln scanners. Maybe we should add descriptions for the pass/warning/dangerCount fields too

docs/operator/getting-started.md
```diff
@@ -2,34 +2,67 @@ Assuming that you installed the operator in the `starboard-operator` namespace,
...
+kubectl create deployment nginx --image nginx:1.16
+```
 
-When the first pod controlled by the `nginx` Deployment is created, the operator
-immediately detects that and creates the Kubernetes job in the
-`starboard-operator` namespace to scan the `nignx` container's image (`nginx:1.16`)
-for vulnerabilities.
+When the first pod controlled by the `nginx` Deployment is started, the operator
+immediately detects that and creates the Kubernetes job in the `starboard-operator`
+namespace to scan the `nignx` container's image (`nginx:1.16`) for
```
lizrice, line 11:
```suggestion
namespace to scan the `nginx` container's image (`nginx:1.16`) for
```

docs/operator/getting-started.md
```diff
@@ -2,34 +2,67 @@ Assuming that you installed the operator in the `starboard-operator` namespace,
...
+scan-configauditreport-c4956cb9d     0/1           1s         1s
+scan-vulnerabilityreport-c4956cb9d   0/1           1s         1s
+```
 
-In our case you should see only one job with a random name scheduled to scan
-the `nginx` Deployment:
+If everything goes fine, the scan jobs are deleted, and the operator saves scan
+reports as custom resources in the `default` namespace. The instance of the
+`vulnerabilityreports` resource named after the active ReplicaSet of the `nginx`
+Deployment is a vulnerability report for the `nginx` container:
```
lizrice, line 25: I found this sentence a bit confusing, how about something like this? 

```suggestion
reports as custom resources in the `default` namespace, named after the Deployment's active ReplicaSet. For image vulnerability scans, the operator creates a report for each different container defined in the active ReplicaSet. In this example there is just one container image called `nginx`:
```

docs/operator/getting-started.md
```diff
@@ -2,34 +2,67 @@ Assuming that you installed the operator in the `starboard-operator` namespace,
...
+NAME                                REPOSITORY      TAG    SCANNER   AGE     CRITICAL   HIGH   MEDIUM   LOW   UNKNOWN
+replicaset-nginx-7ff78f74b9-nginx   library/nginx   1.16   Trivy     9m57s   1          33     22       86    0
+```
 
-If everything goes fine, the scan job is deleted, and the operator creates the
-`vulnerabilityreports` resource in the `default` namespace that corresponds to
-the `nginx` container:
+Similarly, the instance of the `configauditreports` resource holds the result
+of auditing the configuration of the active ReplicaSet controlled by the `nginx`
+Deployment:
```
lizrice, line 35:
```suggestion
Similarly, the operator creates a `configauditreport` resource holding the result
of auditing the configuration of the active ReplicaSet controlled by the `nginx`
Deployment:
```

docs/operator/installation/olm.md
```diff
@@ -5,88 +5,97 @@ You can install the Starboard operator from [OperatorHub.io](https://operatorhub
...
+   ```yaml
+   apiVersion: operators.coreos.com/v1alpha2
+   kind: OperatorGroup
+   metadata:
+     name: starboard-operator
+     namespace: starboard-operator
+   spec:
+     targetNamespaces:
+     - default
+   ```
```
lizrice, line 29: I think this is missing the `kubectl apply` command to apply this?

docs/operator/installation/olm.md
```diff
@@ -5,88 +5,97 @@ You can install the Starboard operator from [OperatorHub.io](https://operatorhub
...
+     installPlanApproval: Automatic
+     config:
+       env:
+       - name: OPERATOR_SCAN_JOB_TIMEOUT
+         value: "60s"
+       - name: OPERATOR_CONCURRENT_SCAN_JOBS_LIMIT
+         value: "10"
+       - name: OPERATOR_LOG_DEV_MODE
+         value: "true"
+   ```
```
lizrice, line 65: Similarly also need to apply this YAML

deploy/init/01-starboard.ns.yaml
```diff
@@ -1,7 +0,0 @@
-apiVersion: v1
```
lizrice, line 1: I didn't check into why it's OK to delete these YAML files, but I imagine you know what you're doing :-)

lizrice (approved):
Couple of suggestions to look at but bascially LGTM!

deploy/crd/configauditreports.crd.yaml
danielpacak: Added descriptions for all columns

docs/operator/installation/olm.md
danielpacak: Good catch. Added back

docs/operator/installation/olm.md
danielpacak: Added back

deploy/init/01-starboard.ns.yaml
danielpacak: :) Basically it's a duplicate of what we already have in deploy/static/*.yaml

PR #359, opened 2021-01-23
Set security context for trivy

mlevesquedion:
This PR adds SecurityContexts to the `trivy` Job in order to run the job with the least amount of privilege possible, in partial fulfillment of #163.

I ran the job with and without the changes in a local kind cluster and I did not see a difference in the output.

To obtain the output, I ran:
```
kubectl create deployment nginx --image nginx:1.16
kubectl wait --for=condition=available deploy/nginx --timeout=60s
./bin/starboard scan vulnerabilityreports deployment/nginx
./bin/starboard get vulnerabilities deployment/nginx -o yaml >vulns.yaml
```

---

Looking at issues on the `trivy` repo, I found this issue: https://github.com/aquasecurity/trivy/issues/580, where running as a non-root user was causing `trivy` to be unable to run. It seems that this was due to using the Docker socket, which is not the case in `starboard`, so I don't think this could cause problems here.

### DIFF
--- a/pkg/trivy/plugin.go
+++ b/pkg/trivy/plugin.go
@@ -250,6 +250,14 @@ func (s *scanner) getPodSpecForStandaloneMode(spec corev1.PodSpec, credentials m
 					MountPath: "/var/lib/trivy",
 				},
 			},
+			SecurityContext: &corev1.SecurityContext{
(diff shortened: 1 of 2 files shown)


danielpacak (approved):
Thank you @mlevesquedion for taking care of this one! LGTM

PR #360, opened 2021-01-23
Set security context for polaris

mlevesquedion:
This PR adds SecurityContexts to the `polaris` Job in order to run the job with the least amount of privilege possible, in partial fulfillment of #163.

I ran the job with and without the changes in a local kind cluster and I did not see a difference in the output.

To obtain the output, I ran:
```
kubectl create deployment nginx --image nginx:1.16
kubectl wait --for=condition=available deploy/nginx --timeout=60s
./bin/starboard scan configauditreports deployment/nginx
./bin/starboard get configaudit deployment/nginx -o yaml >cars.yaml
```


danielpacak (approved):
LGTM! Thank you @mlevesquedion !

PR #361, opened 2021-01-25
test(operator): Test that ConfigAuditReport is created

danielpacak:

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -109,12 +109,16 @@ jobs:
           kubectl describe node
       - name: Run integration tests
         run: |
+          kubectl apply -f deploy/static/01-starboard-operator.ns.yaml \
+            -f deploy/static/02-starboard-operator.sa.yaml \
+            -f deploy/static/03-starboard-operator.clusterrole.yaml \
+            -f deploy/static/04-starboard-operator.clusterrolebinding.yaml
           make itests-starboard-operator
         env:
           KUBECONFIG: /home/runner/.kube/config
-          OPERATOR_NAMESPACE: default
+          OPERATOR_NAMESPACE: starboard-operator
+          OPERATOR_SERVICE_ACCOUNT: starboard-operator
           OPERATOR_TARGET_NAMESPACES: default
-          OPERATOR_SERVICE_ACCOUNT: ""
       - name: Upload code coverage
         uses: codecov/codecov-action@v1
         with:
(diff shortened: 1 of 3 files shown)


.github/workflows/build.yaml
```diff
@@ -109,12 +109,16 @@ jobs:
           kubectl describe node
       - name: Run integration tests
         run: |
+          kubectl apply -f deploy/static/01-starboard-operator.ns.yaml \
```
danielpacak, line 112: We're running the operator as cluster admin out of cluster to generate test coverage report. However, scan jobs are run with `starboard-operator` service account. Therefore, proper RBAC config is required before we run the test suite.

itest/starboard-operator/starboard_operator_test.go
```diff
@@ -15,55 +15,80 @@ import (
...
-				},
-				Spec: appsv1.DeploymentSpec{
-					Replicas: pointer.Int32Ptr(1),
-					Selector: &metav1.LabelSelector{
-						MatchLabels: map[string]string{"app": "wordpress"},
+	Describe("When a new Deployment is created", func() {
+
+		BeforeEach(func() {
+			_, err := kubeClientset.AppsV1().Deployments(namespaceName).
+				Create(context.TODO(), &appsv1.Deployment{
```
knqyf263, line 33: nit: What if you define `ctx` at first and use it for each method? `context.TODO` makes me feel like you are planning to pass the upstream `ctx` in the future.

```suggestion
	Describe("When a new Deployment is created", func() {
		ctx := context.Background()
		BeforeEach(func() {
			_, err := kubeClientset.AppsV1().Deployments(namespaceName).
				Create(ctx, &appsv1.Deployment{
```

>TODO returns a non-nil, empty Context. Code should use context.TODO when it's unclear which Context to use or it is not yet available (because the surrounding function has not yet been extended to accept a Context parameter).

https://golang.org/pkg/context/#TODO

In this case, it doesn't seem that the same `ctx` needs to be shared, so you can just pass `context.Background()`.

```suggestion
	Describe("When a new Deployment is created", func() {

		BeforeEach(func() {
			_, err := kubeClientset.AppsV1().Deployments(namespaceName).
				Create(context.Background(), &appsv1.Deployment{
```

knqyf263 (approved):

itest/starboard-operator/starboard_operator_test.go
danielpacak: That's a good point. Agreed.

PR #370, opened 2021-01-29
fix: Save security reports for static pods

danielpacak:
Note that kublet can manage pods independently by reading pod definition
files from a configured host directory (typically /etc/kubernetes/manifests).
Such pods are called *static pods* and are owned by a cluster Node.
So far we could not save vulnerability and config audit reports for such pods.

In this commit we treat them as unmanaged pods. (Otherwise we'd require
cluster-scoped permissions to get Nodes in order to set the owner reference
when we create an instance of custom resource.)

Resolves: #234

### DIFF
--- a/pkg/resources/resources.go
+++ b/pkg/resources/resources.go
@@ -36,12 +36,32 @@ func GetContainerImagesFromJob(job *batchv1.Job) (kube.ContainerImages, error) {
 	return containerImages, nil
 }
 
-// GetImmediateOwnerReference returns the immediate owner of the specified Pod.
-// For example, for a Pod controlled by a Deployment it will return the active ReplicaSet object,
-// whereas for an unmanaged Pod the immediate owner is the Pod itself.
+// GetImmediateOwnerReference returns the immediate owner of the specified pod.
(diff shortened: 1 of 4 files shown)


pkg/resources/resources.go
```diff
@@ -36,21 +36,41 @@ func GetContainerImagesFromJob(job *batchv1.Job) (kube.ContainerImages, error) {
...
 
-// GetImmediateOwnerReference returns the immediate owner of the specified Pod.
-// For example, for a Pod controlled by a Deployment it will return the active ReplicaSet object,
-// whereas for an unmanaged Pod the immediate owner is the Pod itself.
+// GetImmediateOwnerReference returns the immediate owner of the specified pod.
+// For example, for a pod controlled by a Deployment it will return the active
+// ReplicaSet, whereas for an unmanaged pod the immediate owner is the pod
+// itself.
+//
+// Note that kublet can manage pods independently by reading pod definition
```
lizrice, line 44:
```suggestion
// Note that kubelet can manage pods independently by reading pod definition
```

lizrice (approved):
I haven't tested it but the approach sounds right, and LGTM

Issue #373, opened 2021-02-01
Manage pods controlled by Argo Workflow


danielpacak:
Pods created by [Argo] controllers are owned by Workflow objects. In this case Starboard fails to resolve the owner reference and cannot save a security report in etcd.

Pods controlled by Workflows or any other non-standard controller should be treated as unmanaged pods. Even though we could read an instance of Workflow in a generic way and set it as the owner of a VulnerabilityReport, this would require additional RBAC config, which depends on third-party software running in the cluster.

```yaml
metadata:
  ownerReferences:
  - apiVersion: argoproj.io/v1alpha1
    blockOwnerDeletion: true
    controller: true
    kind: Workflow
    name: hello-world-r99sq
    uid: 7791f1e4-9486-4b38-a09f-82806b57c4ae
```

[Argo]: https://github.com/argoproj/argo/

PR #376, opened 2021-02-03
feat: Manage pods controlled by third parties

danielpacak:
Pods created and controlled by third party frameworks,
such as Argo workflow engine, are considered as unmanaged.
Otherwise we'd need to maintain and extend the list of
RBAC permissions over time.

Resolves: #373

### DIFF
--- a/pkg/resources/resources.go
+++ b/pkg/resources/resources.go
@@ -48,26 +48,25 @@ func GetContainerImagesFromJob(job *batchv1.Job) (kube.ContainerImages, error) {
 // cluster-scoped permissions to get Nodes in order to set the owner reference
 // when we create an instance of custom security report.)
 //
+// Pods created and controlled by third party frameworks, such as Argo workflow
+// engine, are considered as unmanaged. Otherwise we'd need to maintain and
+// extend the list of RBAC permissions over time.
 // TODO Merge this method with OwnerResolver, which accepts kube.Object and resolves client.Object.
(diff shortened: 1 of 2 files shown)


pkg/resources/resources.go
```diff
@@ -48,26 +48,38 @@ func GetContainerImagesFromJob(job *batchv1.Job) (kube.ContainerImages, error) {
...
-			return kube.Object{
-				Kind:      kube.KindPod,
-				Namespace: pod.Namespace,
-				Name:      pod.Name,
-			}
+	if ownerRef == nil {
+		return kube.Object{
+			Kind:      kube.KindPod,
+			Namespace: pod.Namespace,
+			Name:      pod.Name,
```
lizrice, line 62: This is super picky, but I don't love that the unmanaged pod version appears 3 times! Could be something like

```
if ownerRef != nil {
  switch ownerRef.Kind {
    case "Pod", "ReplicaSet"... :
      return kube.Object {
        Kind:      kube.Kind(ownerRef.Kind),
        ...
      }
  }
}

// Pod owned by anything else is treated the same as an unmanaged pod
return kube.Object {
        Kind: kube.KindPod,
        ...
}
```

lizrice (approved):
Suggestion

pkg/resources/resources.go
danielpacak: You're right. I've applied your suggestion. Reads much better now.

PR #388, opened 2021-02-12
Make kubehunter quick config param optional

mlevesquedion:
This PR addresses a potential issue with the `kube-hunter.quick` configuration param. It doesn't make much sense for this parameter to be required, since it's a fairly specific setting that a significant portion of users may not care about.

(This PR was inspired by https://github.com/aquasecurity/starboard/discussions/386.)


danielpacak (approved):
Thanks @mlevesquedion LGTM!

PR #389, opened 2021-02-12
Get vulnerabilities from replicaset in same hierarchy

### LINKED ISSUE #339: [cli] The get vulnerabilities command should resolve reports owned by any object in the same workload hierarchy
If a VulnerabilityReport is owned by ReplicaSet (the default behaviour of the operator) you cannot retrieve it with starboard CLI by specifying the Deployment as a command-line arg

```
starboard get vulnerabilities deploy/nginx
```

If the VulnerabilityReport is not owned by a Deployment, the implementation should lookup the active ReplicaSet / revision and then find the report.

Similarly, when you specify a pod name as command-line argument it should lookup its parent ReplicaSet and find the report at ReplicaSet level.

EDIT: Related to that is to think about the interface that we currently provide for security reports, which is somehow mixing what `kubectl get` and `kubectl describe` commands do for built in resources. In other words, we should consider adding `starboard describe <report type> <workload>` command.

mlevesquedion:
Fixes #339.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -26,7 +26,7 @@ on:
 env:
   GO_VERSION: "1.15"
   KIND_VERSION: "v0.9.0"
-  KIND_IMAGE: "kindest/node:v1.18.8@sha256:f4bcc97a0ad6e7abaf3f643d890add7efe6ee4ab90baeb374b4f41a4c95567eb"
+  KIND_IMAGE: "kindest/node:v1.19.7@sha256:a70639454e97a4b733f9d9b67e12c01f6b0297449d5b9cbbef87473458e26dca"
 jobs:
   release-snapshot:
     name: Release unversioned snapshot
--- a/pkg/vulnerabilityreport/io.go
+++ b/pkg/vulnerabilityreport/io.go
@@ -10,6 +10,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/kubernetes"
 	"k8s.io/klog"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
@@ -36,14 +37,16 @@ type ReadWriter interface {
 }
 
 type readWriter struct {
-	clientset versioned.Interface
+	starboardClientset  versioned.Interface
+	kubernetesClientset kubernetes.Interface
 }
 
 // NewReadWriter constructs a new ReadWriter which is using the client-go
 // module for interacting with the Kubernetes API server.
-func NewReadWriter(clientset versioned.Interface) ReadWriter {
+func NewReadWriter(starboardClientset versioned.Interface, kubernetesClientset kubernetes.Interface) ReadWriter {
 	return &readWriter{
-		clientset: clientset,
+		starboardClientset:  starboardClientset,
+		kubernetesClientset: kubernetesClientset,
 	}
 }
 
@@ -58,7 +61,7 @@ func (r *readWriter) Write(ctx context.Context, reports []v1alpha1.Vulnerability
 }
 
 func (r *readWriter) createOrUpdate(ctx context.Context, report v1alpha1.VulnerabilityReport) error {
-	existing, err := r.clientset.AquasecurityV1alpha1().VulnerabilityReports(report.Namespace).
+	existing, err := r.starboardClientset.AquasecurityV1alpha1().VulnerabilityReports(report.Namespace).
 		Get(ctx, report.Name, metav1.GetOptions{})
 
 	if err == nil {
@@ -67,14 +70,14 @@ func (r *readWriter) createOrUpdate(ctx context.Context, report v1alpha1.Vulnera
 		deepCopy.Labels = report.Labels
 		deepCopy.Report = report.Report
 
-		_, err = r.clientset.AquasecurityV1alpha1().VulnerabilityReports(report.Namespace).
+		_, err = r.starboardClientset.AquasecurityV1alpha1().VulnerabilityReports(report.Namespace).
 			Update(ctx, deepCopy, metav1.UpdateOptions{})
 		return err
 	}
 
 	if errors.IsNotFound(err) {
 		klog.V(3).Infof("Creating VulnerabilityReport %q", report.Namespace+"/"+report.Name)
-		_, err = r.clientset.AquasecurityV1alpha1().VulnerabilityReports(report.Namespace).
+		_, err = r.starboardClientset.AquasecurityV1alpha1().VulnerabilityReports(report.Namespace).
 			Create(ctx, &report, metav1.CreateOptions{})
 		return err
 	}
@@ -83,7 +86,7 @@ func (r *readWriter) createOrUpdate(ctx context.Context, report v1alpha1.Vulnera
 }
 
 func (r *readWriter) FindByOwner(ctx context.Context, owner kube.Object) ([]v1alpha1.VulnerabilityReport, error) {
-	list, err := r.clientset.AquasecurityV1alpha1().VulnerabilityReports(owner.Namespace).List(ctx, metav1.ListOptions{
+	list, err := r.starboardClientset.AquasecurityV1alpha1().VulnerabilityReports(owner.Namespace).List(ctx, metav1.ListOptions{
 		LabelSelector: labels.Set{
 			kube.LabelResourceKind:      string(owner.Kind),
 			kube.LabelResourceName:      owner.Name,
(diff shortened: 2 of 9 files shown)


.github/workflows/build.yaml
```diff
@@ -26,7 +26,7 @@ on:
 env:
   GO_VERSION: "1.15"
   KIND_VERSION: "v0.9.0"
-  KIND_IMAGE: "kindest/node:v1.18.8@sha256:f4bcc97a0ad6e7abaf3f643d890add7efe6ee4ab90baeb374b4f41a4c95567eb"
+  KIND_IMAGE: "kindest/node:v1.19.7@sha256:a70639454e97a4b733f9d9b67e12c01f6b0297449d5b9cbbef87473458e26dca"
```
mlevesquedion, line 29: The below test case was failing in the CI with v1.18.8, but passing locally:
Command get vulnerabilities
 for deployment/nginx resource
  vulnerabilities are associated with the deployment itself
   vulnerabilities are associated with the managed replicaset
    getting vulnerabilities by pod name

I am unable to create a local kind cluster using the v1.18.8 version, so I can't reproduce the failure. I am not sure why the kind version would have an influence on whether or not the test passes, but using the v1.19.7 version, the test passes both locally and in the CI.

danielpacak: That's interesting. Have you tried creating a Kind cluster with a custom config where you could specify the version of Kubernetes?

```yaml
# three node (two workers) cluster config
kind: Cluster
apiVersion: kind.x-k8s.io/v1alpha4
nodes:
- role: control-plane
  image: kindest/node:v1.18.8@sha256:f4bcc97a0ad6e7abaf3f643d890add7efe6ee4ab90baeb374b4f41a4c95567eb
- role: worker
  image: kindest/node:v1.18.8@sha256:f4bcc97a0ad6e7abaf3f643d890add7efe6ee4ab90baeb374b4f41a4c95567eb
- role: worker
  image: kindest/node:v1.18.8@sha256:f4bcc97a0ad6e7abaf3f643d890add7efe6ee4ab90baeb374b4f41a4c95567eb
```

pkg/vulnerabilityreport/io.go
```diff
@@ -93,9 +96,68 @@ func (r *readWriter) FindByOwner(ctx context.Context, owner kube.Object) ([]v1al
...
+	}
+
 	return list.DeepCopy().Items, nil
 }
 
+// findRelatedReplicaset attempts to find the replicaset that is associated with
+// the given owner. If the owner is a Deployment, it will look for a replicaset
+// that is controlled by the Deployment. If the owner is a Pod, it will look for
+// the replicaset that owns the Pod.
+func (r *readWriter) findRelatedReplicaset(ctx context.Context, owner kube.Object) (rsName string, found bool) {
```
danielpacak, line 125: I think we'll need a very similar logic to resolve ConfigAuditReports. So I was wondering if this method can be moved out the vulnerabilityreport package and made public.

pkg/vulnerabilityreport/io.go
```diff
@@ -93,9 +96,68 @@ func (r *readWriter) FindByOwner(ctx context.Context, owner kube.Object) ([]v1al
...
+	switch owner.Kind {
+	// find a replicaset controlled by the deployment
+	case kube.KindDeployment:
+		rsList, err := r.kubernetesClientset.AppsV1().ReplicaSets(owner.Namespace).List(ctx, metav1.ListOptions{})
```
danielpacak, line 129: I know that this would require loading Deployment object, but maybe we can use label selector (deployment.Spec.Selector) to retrieve ReplicaSets instead of loading all of them in the specified namespace.

pkg/vulnerabilityreport/io.go
```diff
@@ -93,9 +96,68 @@ func (r *readWriter) FindByOwner(ctx context.Context, owner kube.Object) ([]v1al
...
+		if err != nil {
+			return "", false
```
danielpacak, line 131: What if loading ReplicaSets has failed? I think we should return error instead of saying that the ReplicaSet was not found. There might be temporary connection error, which does not mean that the ReplicaSet is not there.

pkg/vulnerabilityreport/io.go
```diff
@@ -93,9 +96,68 @@ func (r *readWriter) FindByOwner(ctx context.Context, owner kube.Object) ([]v1al
...
+		if len(rsList.Items) == 0 {
+			return "", false
+		}
+		for _, rs := range rsList.Items {
+			if len(rs.OwnerReferences) == 0 {
+				continue
+			}
+			for _, ownerRef := range rs.OwnerReferences {
+				if owner.Name == ownerRef.Name && *ownerRef.Controller {
+					return rs.Name, true
```
danielpacak, line 142: I believe that we want to lookup the active ReplicaSet (aka current revision) here. For that I'd probably lookup and compare the `deployment.kubernetes.io/revision` annotation set on both Deployment and the ReplicaSet objects.

PS We're doing sth similar in Octant plugin [here](https://github.com/aquasecurity/starboard-octant-plugin/blob/main/pkg/plugin/model/repository.go#L116). However, the code cannot be shared easily as Octant has it's own Kubernetes client interfaces.

pkg/vulnerabilityreport/io.go
```diff
@@ -93,9 +96,68 @@ func (r *readWriter) FindByOwner(ctx context.Context, owner kube.Object) ([]v1al
...
+				}
+			}
+		}
+	// find the replicaset managing the pod
+	case kube.KindPod:
```
danielpacak, line 147: What do you think about splitting those two cases into two methods GetActiveReplicaSetByDeployment and GetReplicaSetByPod. Reading the code and testing will be simpler.

pkg/vulnerabilityreport/io.go
```diff
@@ -93,9 +96,68 @@ func (r *readWriter) FindByOwner(ctx context.Context, owner kube.Object) ([]v1al
...
+		pod, err := r.kubernetesClientset.CoreV1().Pods(owner.Namespace).Get(ctx, owner.Name, metav1.GetOptions{})
+		if err != nil {
+			return "", false
```
danielpacak, line 150: It's not true in case of Kubernetes API failure / downtime.

pkg/vulnerabilityreport/io.go
```diff
@@ -93,9 +96,68 @@ func (r *readWriter) FindByOwner(ctx context.Context, owner kube.Object) ([]v1al
...
+		}
+		for _, ownerRef := range pod.OwnerReferences {
```
danielpacak, line 152: With the pod instance available you could use the utility `metav1.ControllerOf()`

danielpacak (changes requested):
Thank you for working on this PR @mlevesquedion  I left a few comments to revisit the implementation of ReplicaSet lookup. In the second round I'll focus on reviewing unit / integration tests.

.github/workflows/build.yaml
mlevesquedion: Yeah, when I try with that version of the `kindest/node` image locally, I get:

```
ERROR: failed to create cluster: failed to init node with kubeadm: command "docker exec --privileged kind-control-plane kubeadm init --skip-phases=preflight --config=/kind/kubeadm.conf --skip-token-print --v=6" failed with error: exit status 1
```

pkg/vulnerabilityreport/io.go
mlevesquedion: SG. Done.

pkg/vulnerabilityreport/io.go
mlevesquedion: I think that's a great idea, likely more efficient if there's a very large number of ReplicaSets.

pkg/vulnerabilityreport/io.go
mlevesquedion: Done.

pkg/vulnerabilityreport/io.go
mlevesquedion: Done. I would expect that checking `*ownerRef.Controller` and comparing revisions would give the same result, but I haven't verified this, so let's use revisions since that seems to work in starboard-octant-plugin.

pkg/vulnerabilityreport/io.go
mlevesquedion: Done.

pkg/vulnerabilityreport/io.go
mlevesquedion: Done, returning error now.

pkg/vulnerabilityreport/io.go
mlevesquedion: I didn't know about `ControllerOf`, thanks! Done.

danielpacak (approved):
LGTM! Thank you @mlevesquedion !

PR #394, opened 2021-02-15
Set security context for kube-hunter

mlevesquedion:
Following discussion on #163, this PR adds a security context for `kube-hunter`, but only if the `kube-hunter` version is at least `0.4.1` (or `latest`, because that should only point to `0.4.1` or newer). Unfortunately, this means that if the version is set using the SHA, then no security context will be used.


pkg/kubehunter/scanner.go
```diff
@@ -110,6 +111,33 @@ func (s *Scanner) prepareKubeHunterJob() (*batchv1.Job, error) {
...
+	if isAtLeast(ver, "0.4.1") || ver == "latest" {
+		podSecurityContext = &corev1.PodSecurityContext{
+			RunAsUser:  pointer.Int64Ptr(1000),
+			RunAsGroup: pointer.Int64Ptr(1000),
+			SeccompProfile: &corev1.SeccompProfile{
+				Type: corev1.SeccompProfileTypeRuntimeDefault,
+			},
+		}
+		containerSecurityContext = &corev1.SecurityContext{
+			Privileged:               pointer.BoolPtr(false),
```
danielpacak, line 132: For some reason, with this settings, kube-hunter no longer reports the **CAP_NET_RAW Enabled** vulnerability. I observed that with Kubernetes v1.18.8

danielpacak (commented):
Overall the code looks good, but I have noticed change in behaviour were kube-hunter does not return certain vulnerabilities.

pkg/kubehunter/scanner.go
mlevesquedion: I think this is because `kube-hunter` reports its own capabilities. When deployed via `starboard`, `kube-hunter` runs as a pod, which means it runs the following check:

```py
@handler.subscribe(RunningAsPodEvent)
class PodCapabilitiesHunter(Hunter):
    """Pod Capabilities Hunter
    Checks for default enabled capabilities in a pod
    """

    def execute(self):
        if self.check_net_raw():
            self.publish_event(CapNetRawEnabled())

    def check_net_raw(self):
        logger.debug("Passive hunter's trying to open a RAW socket")
        try:
            # trying to open a raw socket without CAP_NET_RAW will raise PermissionsError
            s = socket.socket(socket.AF_INET, socket.SOCK_RAW, socket.IPPROTO_RAW)
            s.close()
            logger.debug("Passive hunter's closing RAW socket")
            return True
        except PermissionError:
            logger.debug("CAP_NET_RAW not enabled")
```

With the security context in this PR, `CAP_NET_RAW` is disabled, so the pod no longer reports it.

Is this ok?

danielpacak: Hm I got it. This is tricky to decide, but I tend to say that we should generate results similar to what kube-hunter run as standalone program would return. In this case, let's assume that accuracy is more important than the least privileged principle.

mlevesquedion: Sounds good. This check feels a little weird to me since `kube-hunter` is reporting on its own capabilities. So I guess this could help users detect that e.g. they don't have a policy that blocks pods running with `CAP_NET_RAW`, but on the other hand, if users have such a policy, then the pod won't run at all.

An unfortunate side effect of wanting to run with `CAP_NET_RAW` is that the container needs to run as root. Indeed, adding `NET_RAW` in the security context only adds the capability to the inherited and bound sets, but not the effective set, which is what determines whether the container can open a raw socket or not. Running as root is one way to get the effective capability. I don't know other ways that don't require changes to `kube-hunter` itself (but I may be missing something).

I've made the requested change. When running locally in a `kind` cluster, the output is the same with and without the security context.

danielpacak: Let's move on and merge it like that. We can always refine it in the next iteration and / or propose changes to Kube-Hunter.

danielpacak (approved):

PR #404, opened 2021-02-17
feat(operator): Integrate kube-bench

danielpacak:
Resolves: #263

### DIFF
--- a/deploy/static/06-starboard-operator.deployment.yaml
+++ b/deploy/static/06-starboard-operator.deployment.yaml
@@ -43,13 +43,15 @@ spec:
             - name: OPERATOR_LOG_DEV_MODE
               value: "false"
             - name: OPERATOR_CONCURRENT_SCAN_JOBS_LIMIT
-              value: "3"
+              value: "10"
             - name: OPERATOR_SCAN_JOB_RETRY_AFTER
               value: "30s"
             - name: OPERATOR_METRICS_BIND_ADDRESS
               value: ":8080"
             - name: OPERATOR_HEALTH_PROBE_BIND_ADDRESS
               value: ":9090"
+            - name: OPERATOR_CIS_KUBERNETES_BENCHMARK_ENABLED
+              value: "true"
           ports:
             - name: metrics
               containerPort: 8080
--- a/pkg/kubebench/scanner.go
+++ b/pkg/kubebench/scanner.go
@@ -121,7 +121,7 @@ func (s *Scanner) Scan(ctx context.Context, node corev1.Node) (v1alpha1.CISKubeB
 		Report: output,
 	}
 
-	err = controllerutil.SetOwnerReference(&node, &report, s.scheme)
+	err = controllerutil.SetControllerReference(&node, &report, s.scheme)
 	if err != nil {
 		return v1alpha1.CISKubeBenchReport{}, err
 	}
--- a/pkg/operator/controller/ciskubebenchreport.go
+++ b/pkg/operator/controller/ciskubebenchreport.go
@@ -1,0 +1,300 @@
+package controller
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/aquasecurity/starboard/pkg/apis/aquasecurity/v1alpha1"
+	"github.com/aquasecurity/starboard/pkg/kube"
+	"github.com/aquasecurity/starboard/pkg/kubebench"
+	"github.com/aquasecurity/starboard/pkg/operator/etc"
+	. "github.com/aquasecurity/starboard/pkg/operator/predicate"
+	"github.com/go-logr/logr"
+	batchv1 "k8s.io/api/batch/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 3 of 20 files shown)


deploy/static/06-starboard-operator.deployment.yaml
```diff
@@ -43,13 +43,15 @@ spec:
             - name: OPERATOR_LOG_DEV_MODE
               value: "false"
             - name: OPERATOR_CONCURRENT_SCAN_JOBS_LIMIT
-              value: "3"
+              value: "10"
```
danielpacak, line 46: We're running 3 different scanners now, so I think it's worth bumping up the default limit

pkg/kubebench/scanner.go
```diff
@@ -121,7 +121,7 @@ func (s *Scanner) Scan(ctx context.Context, node corev1.Node) (v1alpha1.CISKubeB
 		Report: output,
 	}
 
-	err = controllerutil.SetOwnerReference(&node, &report, s.scheme)
+	err = controllerutil.SetControllerReference(&node, &report, s.scheme)
```
danielpacak, line 124: By changing owner to controller we can take advantage of controller-runtime's controller builder. Such builder can watch the node, but also owned report. Hence, whenever a CISKubeBenchReport is deleted a reconciliation for the owning node is triggered.

pkg/operator/controller/ciskubebenchreport.go
```diff
@@ -0,0 +1,300 @@
...
+			},
+		},
+	}, nil
+}
+
+func (r *CISKubeBenchReportReconciler) getScanJobName(node *corev1.Node) string {
+	return "scan-kubebenchreports-" + node.Name
+}
+
+func (r *CISKubeBenchReportReconciler) reconcileJobs() reconcile.Func {
```
danielpacak, line 191: This code is very similar to other reconcilers, but I'll refactor it in a separate PR by implementing a template method or similar structure to avoid duplicates.

pkg/operator/controller/ciskubebenchreport.go
```diff
@@ -0,0 +1,300 @@
...
+		if err != nil {
+			if errors.IsNotFound(err) {
+				log.V(1).Info("Ignoring job read from cache that must have been deleted")
+				return ctrl.Result{}, nil
+			}
+			return ctrl.Result{}, fmt.Errorf("getting job from cache: %w", err)
+		}
+
+		if len(job.Status.Conditions) == 0 {
+			log.V(1).Info("Job has no conditions despite using predicate")
```
danielpacak, line 206: I found in tests that even though we use predicate, this condition does not always hold. Therefore, I'm double checking here. Very defensive programming, otherwise we'll have panics.

pkg/operator/etc/config.go
```diff
@@ -9,15 +9,16 @@ import (
...
+	Namespace                     string        `env:"OPERATOR_NAMESPACE"`
+	TargetNamespaces              string        `env:"OPERATOR_TARGET_NAMESPACES"`
+	ServiceAccount                string        `env:"OPERATOR_SERVICE_ACCOUNT" envDefault:"starboard-operator"`
+	ScanJobTimeout                time.Duration `env:"OPERATOR_SCAN_JOB_TIMEOUT" envDefault:"5m"`
+	ConcurrentScanJobsLimit       int           `env:"OPERATOR_CONCURRENT_SCAN_JOBS_LIMIT" envDefault:"10"`
+	ScanJobRetryAfter             time.Duration `env:"OPERATOR_SCAN_JOB_RETRY_AFTER" envDefault:"30s"`
+	MetricsBindAddress            string        `env:"OPERATOR_METRICS_BIND_ADDRESS" envDefault:":8080"`
+	HealthProbeBindAddress        string        `env:"OPERATOR_HEALTH_PROBE_BIND_ADDRESS" envDefault:":9090"`
+	LogDevMode                    bool          `env:"OPERATOR_LOG_DEV_MODE" envDefault:"false"`
+	CISKubernetesBenchmarkEnabled bool          `env:"OPERATOR_CIS_KUBERNETES_BENCHMARK_ENABLED" envDefault:"true"`
```
danielpacak, line 21: It's just this new config param added here

pkg/operator/operator.go
```diff
@@ -50,24 +52,29 @@ func Run(buildInfo starboard.BuildInfo, operatorConfig etc.Config) error {
 	case etc.SingleNamespace:
 		// Add support for SingleNamespace set in OPERATOR_NAMESPACE (e.g. `starboard-operator`)
 		// and OPERATOR_TARGET_NAMESPACES (e.g. `default`).
-		cachedNamespaces := append(targetNamespaces, operatorNamespace)
+		cachedNamespaces := append(targetNamespaces, operatorNamespace, "")
```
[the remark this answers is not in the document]
danielpacak: Good catch. It's a bug. We want to add blank namespace only if the CIS Kubernetes Benchmark reconciler is enabled in order to get, list, and watch Nodes, which are cluster-scoped

knqyf263 (approved):
LGTM👍

PR #405, opened 2021-02-18
refactor: Separate kube-bench -specific code

danielpacak:


pkg/kubebench/scanner.go
```diff
@@ -23,41 +25,44 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 )
 
-const (
-	kubeBenchContainerName = "kube-bench"
-)
+// Plugin defines the interface between Starboard and Kubernetes configuration
+// checker with CIS Kubernetes Benchmarks.
+type Plugin interface {
```
danielpacak, line 30: Even though there's (and will be) only one implementation that relies on Kube-Bench, the purpose of this interface is to share the logic between CLI and the operator.

knqyf263 (approved):
LGTM 👍

pkg/kubebench/scanner_test.go
```diff
@@ -0,0 +1,221 @@
...
+				Image:                    "docker.io/aquasec/kube-bench:0.5.0",
+				ImagePullPolicy:          corev1.PullIfNotPresent,
+				TerminationMessagePolicy: corev1.TerminationMessageFallbackToLogsOnError,
+				Command:                  []string{"sh"},
+				Args:                     []string{"-c", "kube-bench --json 2> /dev/null"},
+				SecurityContext: &corev1.SecurityContext{
+					Privileged:               pointer.BoolPtr(false),
+					AllowPrivilegeEscalation: pointer.BoolPtr(false),
+					Capabilities: &corev1.Capabilities{
+						Drop: []corev1.Capability{"all"},
```
joebowbeer, line 108:
```suggestion
						Drop: []corev1.Capability{"ALL"},
```

Isn't this supposed to be ALL (caps)?

According to [spec](https://kubernetes.io/docs/concepts/security/pod-security-standards/#restricted), PSA implementation, and most policy rules?

Issue #395, opened 2021-02-16
Add AVD reference to KubeHunterReport CR


danielpacak:
Kube-Hunter returns avc_reference property so you can learn about detected vulnerability from AVD, e.g. https://avd.aquasec.com/kube-hunter/khv002/

We should extend the kubehunterreports CR to incorporate the AVD links.

krol3:
Nice issue, I'd like to help

PR #407, opened 2021-02-20
chore: add avd_reference to kube-hunter report

krol3:
Related the issue [395](https://github.com/aquasecurity/starboard/issues/395)
Is it necessary to update the lens and octant plugin too?


pkg/apis/aquasecurity/v1alpha1/kube_hunter_types.go
```diff
@@ -63,8 +65,91 @@ var (
...
+																{Raw: []byte(strconv.Quote(string(KubeHunterSeverityHigh)))},
+																{Raw: []byte(strconv.Quote(string(KubeHunterSeverityMedium)))},
+																{Raw: []byte(strconv.Quote(string(KubeHunterSeverityLow)))},
+																{Raw: []byte(strconv.Quote(string(KubeHunterSeverityUnknown)))},
+															},
+														},
+														"vulnerability": {Type: "string"},
+														"description":   {Type: "string"},
+														"evidence":      {Type: "string"},
+														"avdReference":  {Type: "string"},
```
danielpacak, line 145: KubeHunter is using `avd_reference` not `avdReference`. If you provide Open API spec only defined properties are persisted in CRD/etcd.

```suggestion
														"avd_reference":  {Type: "string"},
```

deploy/crd/kubehunterreports.crd.yaml
```diff
@@ -31,8 +31,94 @@ spec:
...
+                        enum:
+                          - high
+                          - medium
+                          - low
+                          - unknown
+                      description:
+                        type: string
+                      evidence:
+                        type: string
+                      avdReference:
```
danielpacak, line 120:
```suggestion
                      avd_reference:
```

deploy/crd/kubehunterreports.crd.yaml
```diff
@@ -31,8 +31,94 @@ spec:
...
+                    type: object
+                    required:
+                      - location
+                      - vid
+                      - category
+                      - severity
+                      - vulnerability
+                      - description
+                      - evidence
+                      - avdReference
```
danielpacak, line 99:
```suggestion
                      - avd_reference
```

danielpacak (changes requested):
Good job @krol3 I think we just have to rename `avdRefernec` to `avd_reference` so the property is actually persisted in etcd.  Make sure it's the case by running `./bin/starboard scan kubehunterreports` and then inspect the actual report.

krol3:
Hi! @danielpacak  I tested this new version, after the update.

`k get kubehunterreports.aquasecurity.github.io -o yaml | more`

```
    vulnerabilities:
    - avd_reference: https://avd.aquasec.com/kube-hunter/none/
      category: Access Risk
      description: |-
        CAP_NET_RAW is enabled by default for pods.
            If an attacker manages to compromise a pod,
            they could potentially take advantage of this capability to perform network
            attacks on other pods running on the same node
```

danielpacak (approved):

PR #408, opened 2021-02-22
chore(release): remove logout step

knqyf263:
If I understand correctly, `docker/login-action@v1` log out from the Docker registry by default. Am I missing something?
https://github.com/docker/login-action#inputs

image
https://github.com/aquasecurity/starboard/runs/1901817059?check_suite_focus=true

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -112,10 +112,5 @@ jobs:
           args: release --rm-dist
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


danielpacak:
> If I understand correctly, `docker/login-action@v1` log out from the Docker registry by default. Am I missing something?
> https://github.com/docker/login-action#inputs
> 
> image
> https://github.com/aquasecurity/starboard/runs/1901817059?check_suite_focus=true

This is correct. Logout step is redundant now. I used to run the docker login command before we refactored to use Docker's GitHub Action.

danielpacak (approved):

PR #446, opened 2021-03-20
doc: Update cli docs to show the correct command to get configaudit

champness:
Just a minor doc fix: The correct command is `starboard get configaudit`, but the doc said `starboard get configauditreport`. This PR updates the doc to match the command.


danielpacak (approved):
Good catch @champness Thank you!

PR #450, opened 2021-03-23
fix: Skip reconciling Jobs managed by CronJob

danielpacak:


pkg/operator/controller/ciskubebenchreport.go
```diff
@@ -118,6 +118,9 @@ func (r *CISKubeBenchReportReconciler) reconcileNodes() reconcile.Func {
 		log.V(1).Info("Scheduling CIS Kubernetes Benchmark checks")
 		err = r.Client.Create(ctx, job)
 		if err != nil {
+			if errors.IsAlreadyExists(err) {
```
danielpacak, line 121: It might happen that cache is not in synch with Kubernetes API state. In this case instead of returning error, which will retry the reconciliation by putting it back to worker queue, we can just ignore and return.

itaysk (approved):

PR #455, opened 2021-03-24
fix: Polaris results parse failed(#449)

### LINKED ISSUE #449: Polaris results parse failed
1. Run the scan command:
`starboard scan configauditreports pod/nginx`
2. Retrieve the report as bellow:
```bash
$bin/starboard get configaudit pod/nginx
....
report:
  containerChecks: null
  podChecks: null
  scanner:
    name: ""
    vendor: ""
    version: ""
  summary:
    dangerCount: 0
    passCount: 0
    warningCount: 0
  updateTimestamp: null
```
As you see, the report was empty.

I have checked the Polaris  output results:
```
I0322 02:20:26.579189 3832089 request.go:655] Throttling request took 2.708992765s, request: xxxxx
```
There are random logs at the first line instead of json output. 

**Environment:**

- Starboard version (use `starboard version`): latest master build 
- Kubernetes version (use `kubectl version`): v1.19.0
- OS (macOS 10.15, Windows 10, Ubuntu 19.10 etc): Ubuntu 18.04

RolandMa1986:
Resolves #449

**What's in the pull request?**
I have created a wrap of _io.Reader_ that can skips invalid characters. it's a very simple solution without additionnally memory copy/ or type convert. (I tried to use a regex by converting the stream to string, but it seems like not necessary.). 
Please feed free to close the pull request if there's a better way.

### DIFF
--- a/pkg/ext/json.go
+++ b/pkg/ext/json.go
@@ -1,0 +1,37 @@
+package ext
+
+import "io"
+
+type JsonReader struct {
+	reader  io.ReadCloser
+	started bool
+}
+
+// NewJsonReader constructs a new instant of JsonReader, which will try to match a start characator of
+// "{" or "[" as beginning of the json stream.
+func NewJsonReader(reader io.ReadCloser) io.ReadCloser {
+	return &JsonReader{reader: reader}
+}
+
+func (j *JsonReader) Read(p []byte) (n int, err error) {
+	if j.started {
+		return j.reader.Read(p)
+	}
+	var c = make([]byte, len(p))
+	for ; err == nil; n, err = j.reader.Read(c) {
+		i := 0
+		for ; i < n; i++ {
+			if c[i] == '{' || c[i] == '[' {
+				j.started = true
+				copy(p, c[i:])
+				return n - i, err
+			}
+		}
+	}
+
+	return n, err
+}
+
+func (j *JsonReader) Close() error {
+	return j.reader.Close()
+}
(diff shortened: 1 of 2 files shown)


danielpacak (commented):
👋 @RolandMa1986 I think that in order to validate your implementation we'd need to add some unit tests. Otherwise it's hard to tell if you handled different cases where JSON output is cluttered with Polaris logs. I could imagine that such clutter may be added somewhere in the middle of JSON output as well.

Also before doing all that I was wondering if we could just redirect stderr to `/dev/null` by changing the command that's run by the Polaris plugin, i.e. : `sh -c "Polaris <other args> 2> /dev/null`. This way we could leave the default unmarshaller intact. WDYT?

RolandMa1986:
@danielpacak , I agree that UT is needed. I will try to add some test cases. 
Those logs are **stdout** instead  of **stderr** actually. Those logs were written by the client-go library.  
Unfortunately, these logs can't be filtered by log level according to the comments. Whenever you initialized a client, client-go will try to call discovery APIs, and those calls may be throttled. And it will only be written to the start or end.

https://github.com/kubernetes/client-go/blob/db49dde90ff2acfc608c11b8e39abda103f43aaa/rest/request.go#L602

danielpacak:
@RolandMa1986 Are you sure that those logs go to stdout instead of stderr by default? Where is that configured? I believe that throttling is logged with klog, which default to stderr

RolandMa1986:
@danielpacak, your are right, it's stdout. just need to redirect the output

PR #461, opened 2021-03-25
refactor: Use custom Gomega matcher to assert VulnerabilityReports

danielpacak:

### DIFF
--- a/itest/starboard/starboard_cli_test.go
+++ b/itest/starboard/starboard_cli_test.go
@@ -1,6 +1,12 @@
 package starboard
 
 import (
+	. "github.com/aquasecurity/starboard/itest/matcher"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	. "github.com/onsi/gomega/gbytes"
+	. "github.com/onsi/gomega/gstruct"
+
 	"context"
 	"os"
 	"strings"
@@ -10,10 +16,6 @@ import (
 	"github.com/aquasecurity/starboard/pkg/cmd"
 	"github.com/aquasecurity/starboard/pkg/kube"
 	"github.com/aquasecurity/starboard/pkg/starboard"
-	. "github.com/onsi/ginkgo"
-	. "github.com/onsi/gomega"
-	. "github.com/onsi/gomega/gbytes"
-	. "github.com/onsi/gomega/gstruct"
 	appsv1 "k8s.io/api/apps/v1"
 	batchv1 "k8s.io/api/batch/v1"
 	batchv1beta1 "k8s.io/api/batch/v1beta1"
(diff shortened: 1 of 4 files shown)


itest/starboard/starboard_cli_test.go
```diff
@@ -61,6 +64,63 @@ func NewPodSpec(podName string, containers map[string]string, args ...string) (*
...
+	return m.Match(actual)
+}
+
+func (matcher *vulnerabilityReportMatcher) FailureMessage(actual interface{}) (message string) {
+	return matcher.failureMessage
+}
+
+func (matcher *vulnerabilityReportMatcher) NegatedFailureMessage(actual interface{}) (message string) {
+	return matcher.negatedFailureMessage
+}
```
simar7, line 122: Just for my understanding: Where are the values for these fields initialized?

itest/starboard/starboard_cli_test.go
```diff
@@ -61,6 +64,63 @@ func NewPodSpec(podName string, containers map[string]string, args ...string) (*
...
+func (matcher *vulnerabilityReportMatcher) Match(actual interface{}) (success bool, err error) {
+	_, ok := actual.(v1alpha1.VulnerabilityReport)
+	if !ok {
+		return false, fmt.Errorf("vulnerabilityReportMatcher expects a VulnerabilityReport")
+	}
+	gvk, err := apiutil.GVKForObject(matcher.owner, starboard.NewScheme())
+	if err != nil {
+		return false, err
+	}
+	m := MatchFields(IgnoreExtras, Fields{
```
simar7, line 91: For my understanding: What's MatchFields? Where is it defined?

danielpacak: With Ginkgo / Gomega a common approach is to use dot imports - see imports sections:

```go
import (
	. "github.com/onsi/ginkgo"
	. "github.com/onsi/gomega"
	. "github.com/onsi/gomega/gbytes"
	. "github.com/onsi/gomega/gstruct"
)
```

This is syntactic sugar, you could also [avoid dot imports](https://onsi.github.io/ginkgo/#avoiding-dot-imports).

itest/starboard/starboard_cli_test.go
danielpacak: A custom matcher is stateful and we set those values in the `Match(actual interface{})` callback. Lines 110-111.

danielpacak: However, I realised that I assigned messages before calling the `Match` method 🤦 I've already fixed that, but in general it is guaranteed that `FailureMessage` and `NegatedFailureMessage` will only be called after `Match`, so you can save off any state you need to compute the messages in `Match`. See https://onsi.github.io/gomega/#adding-your-own-matchers

itest/starboard/starboard_cli_test.go
simar7: Ah lazy me should've opened the PR in an IDE to realize it was a dot import. That makes a lot more sense!

itest/starboard/starboard_cli_test.go
simar7: I see that makes a lot more sense. Cool!

simar7 (approved):

Issue #485, opened 2021-04-08
Bug in helm chart v0.10.0-rc4: error calling gt: incompatible types for comparison


dschunack:
**What steps did you take and what happened:**
Hi,

we used the helm chart to deploy the version v0.10.0-rc4 on a test k8s, but the new chart is not working correct.
```
executing "starboard-operator/templates/rbac.yaml" at <gt .Values.operator.replicas 1>: error calling gt: incompatible types for comparison
executing "starboard-operator/templates/deployment.yaml" at <gt .Values.operator.replicas 1>: error calling gt: incompatible types for comparison
```
The code of the chart was change with pull request  #452 

**What did you expect to happen:**
Installation without issues as in version v0.9.2

**Anything else you would like to add:**

I created a pull request to fix the issue.  #486 

**Environment:**

- Starboard version (use `starboard version`): 
- Starboard Version: {Version:0.6.0 Commit:ad6b9b99931d7524806f79b1329fdda82c75a715 Date:2020-10-14T20:14:11Z}

- Kubernetes version (use `kubectl version`): 
- Client Version: version.Info{Major:"1", Minor:"20+", GitVersion:"v1.20.4-dirty", GitCommit:"e87da0bd6e03ec3fea7933c4b5263d151aafd07c", GitTreeState:"dirty", BuildDate:"2021-03-15T09:58:13Z", GoVersion:"go1.16.2", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"18+", GitVersion:"v1.18.9-eks-d1db3c", GitCommit:"d1db3c46e55f95d6a7d3e5578689371318f95ff9", GitTreeState:"clean", BuildDate:"2020-10-20T22:18:07Z", GoVersion:"go1.13.15", Compiler:"gc", Platform:"linux/amd64"}

- OS (macOS 10.15, Windows 10, Ubuntu 19.10 etc): macOS 11

PR #486, opened 2021-04-08
Fix helm chart issue: error calling gt: incompatible types for comparison

dschunack:
Hi,

this is the pull request to fix the issues reported in #485


danielpacak (approved):
Good catch @dschunack and thank you for fixing it!

Issue #483, opened 2021-04-08
Conftest plugin shows negative pass count


danielpacak:
```console
$ kubectl get configaudit -A -o wide
NAMESPACE     NAME                                    SCANNER    AGE     PASS   DANGER   WARNING
kube-system   daemonset-gke-metrics-agent             Conftest   22h     23     10       0
kube-system   replicaset-kube-dns-6465f78586          Conftest   22h     -5     38       0
```

PR #488, opened 2021-04-09
fix(conftest): Do not show negative pass count

danielpacak:
Resolves: #483

### DIFF
--- a/pkg/plugin/conftest/plugin.go
+++ b/pkg/plugin/conftest/plugin.go
@@ -65,7 +65,8 @@ func (p *plugin) GetScanJobSpec(ctx starboard.PluginContext, obj client.Object) 
 
 	secret := &corev1.Secret{
 		ObjectMeta: metav1.ObjectMeta{
-			Name: p.idGenerator.GenerateID(),
+			Name:      p.idGenerator.GenerateID(),
+			Namespace: ctx.GetNamespace(),
 		},
 		StringData: map[string]string{
 			"workload.yaml": string(workloadAsYAML),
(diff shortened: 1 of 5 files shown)


pkg/plugin/conftest/plugin.go
```diff
@@ -187,6 +188,9 @@ func (p *plugin) getPolicies(ctx starboard.PluginContext) ([]string, error) {
 		if !strings.HasPrefix(key, "conftest.policy.") {
 			continue
 		}
+		if !strings.HasSuffix(key, ".rego") {
```
danielpacak, line 191: Conftest will skip a policy file without the .rego suffix so there's no point in mounting it

pkg/plugin/conftest/plugin.go
```diff
@@ -210,7 +214,10 @@ func (p *plugin) ParseConfigAuditReportData(logsReader io.ReadCloser) (v1alpha1.
 	var successesCount, warningCount, dangerCount int
 
 	for _, cr := range checkResults {
-		successesCount += cr.Successes
+		// Conftest reportedly returns negative count of passed tests is some cases: https://github.com/open-policy-agent/conftest/issues/464
+		if cr.Successes > 0 {
```
danielpacak, line 218: This is the actual fix. The rest is adding unit tests.

pkg/plugin/polaris/model_test.go
```diff
@@ -0,0 +1 @@
+package polaris_test
```
danielpacak, line 1: To properly measure test coverage

pkg/plugin/conftest/plugin.go
simar7: I read through the conftest bug report and to me it sounds like this happens if there are more failures than passes in a rego file. If we add this check of >0 will we lose anything?

simar7 (approved):

pkg/plugin/conftest/plugin.go
danielpacak: I believe that we are not loosing anything because Conftest does not return successful checks anyway, only warnings, failures and successes count. Successes count as as a negative integer is just confusing.

PR #493, opened 2021-04-12
fix(polaris): Remove clutter from JSON output

danielpacak:
Resolves: #449

### DIFF
--- a/pkg/plugin/polaris/plugin_test.go
+++ b/pkg/plugin/polaris/plugin_test.go
@@ -9,8 +9,7 @@ import (
 	"github.com/aquasecurity/starboard/pkg/ext"
 	"github.com/aquasecurity/starboard/pkg/plugin/polaris"
 	"github.com/aquasecurity/starboard/pkg/starboard"
-	"github.com/stretchr/testify/assert"
-	"github.com/stretchr/testify/require"
+	"github.com/onsi/gomega"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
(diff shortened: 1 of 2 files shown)


pkg/plugin/polaris/plugin_test.go
```diff
@@ -9,8 +9,7 @@ import (
 	"github.com/aquasecurity/starboard/pkg/ext"
 	"github.com/aquasecurity/starboard/pkg/plugin/polaris"
 	"github.com/aquasecurity/starboard/pkg/starboard"
-	"github.com/stretchr/testify/assert"
-	"github.com/stretchr/testify/require"
+	"github.com/onsi/gomega"
```
danielpacak, line 12: On the mission to replace Testify with Gomega matchers, which are more powerful and we use them in integration tests.

pkg/plugin/polaris/plugin_test.go
```diff
@@ -87,12 +86,10 @@ func TestPlugin_GetScanJobSpec(t *testing.T) {
...
 						},
-						Command: []string{"polaris"},
+						Command: []string{"sh"},
 						Args: []string{
-							"audit",
-							"--log-level", "error",
-							"--config", "/etc/starboard/polaris.config.yaml",
-							"--resource", "default/Deployment.apps/v1/nginx",
+							"-c",
+							"polaris audit --log-level error --config /etc/starboard/polaris.config.yaml --resource default/Deployment.apps/v1/nginx 2> /dev/null",
```
danielpacak, line 92: 2> /dev/null to get rid of logs from clinet-go, e.g. rate limit warnings

pkg/plugin/polaris/plugin_test.go
simar7: I get the reasoning but I personally would still vote for testify just because I find them simpler to write. gexec is great though for integration tests.

simar7 (approved):
behold our new saviour gomega

pkg/plugin/polaris/plugin_test.go
danielpacak: With testify I struggled many times to assert on complex objects / object trees. Only Gomega allowed me to do so. On the other hand keeping and learning 2 assert libraries doesn't really make sense for one project. But we can think about it and vote what would be the best option.

Issue #509, opened 2021-04-19
Conftest: configmap references non-existent config key: *****.rego


danielpacak:
**What steps did you take and what happened:**

Modify `starboard-conftest-config` ConfigMap by deleting the `runs_as_root_user.rego` policy after a scan Job was created but before it was actually run. This ends up with with the following error and Pod stuck in the Pending status:

```
MountVolume.SetUp failed for volume "policies" : configmap references non-existent config key: conftest.policy.runs_as_root_user.rego
```

**What did you expect to happen:**

The scan Job has completed even though the ConfigMap was updated.

**Anything else you would like to add:**

N/A

**Environment:**

- Starboard version (use `starboard version`): v0.10.0
- Kubernetes version (use `kubectl version`): any
- OS (macOS 10.15, Windows 10, Ubuntu 19.10 etc): any

PR #511, opened 2021-04-19
fix(Conftest): Error: configmap references non-existent config key

danielpacak:
When we created a scan Job for config audits it referred directly
to the starboard-conftest-config ConfigMap. However, the ConfigMap
might have changed before the Job was run. When we removed a policy
from the ConfigMap the Job got stuck in the Pending state due to
non-existent config key error.

The solution is to copy policies to a temporary Secret and associate
it with the scan Job so we're agnostic to changes of the
starboard-conftest-config ConfigMap.

Resolves: #509


deven0t (approved):
LGTM

PR #516, opened 2021-04-20
fix(Conftest): Do not specify UID (1000) and GID (1000) in SecurityContext

danielpacak:
For upstream K8s cluster scan Jobs will run as USER defined in
the Conftest container image. For OpenShift Container Platform
the admission controller will assign UID and GID based on the
range defined for the OpenShift project / K8s namespace where
Starboard is installed.

### DIFF
--- a/itest/starboard/suite_test.go
+++ b/itest/starboard/suite_test.go
@@ -17,8 +17,6 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-const namespaceItest = "starboard-itest"
-
 var (
 	kubeClient             client.Client
 	apiextensionsClientset apiextensions.ApiextensionsV1beta1Interface
@@ -34,7 +32,7 @@ var (
 	}
 	testNamespace = &corev1.Namespace{
 		ObjectMeta: metav1.ObjectMeta{
-			Name: namespaceItest,
+			Name: "starboard-itest",
 		},
 	}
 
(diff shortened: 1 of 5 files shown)


pkg/plugin/conftest/plugin.go
```diff
@@ -166,13 +166,7 @@ func (p *plugin) GetScanJobSpec(ctx starboard.PluginContext, obj client.Object)
 					},
 				},
 			},
-			SecurityContext: &corev1.PodSecurityContext{
-				RunAsUser:  pointer.Int64Ptr(1000),
```
danielpacak, line 170: - This user is not defined in Conftest image
- On OpenShift Container Platform it blocks scan Job from running

pkg/operator/operator.go
```diff
@@ -26,8 +26,6 @@ var (
 )
 
 func Run(buildInfo starboard.BuildInfo, operatorConfig etc.Config) error {
-	setupLog.Info("Starting operator", "buildInfo", buildInfo)
```
danielpacak, line 29: This is duplicate log entry

itest/starboard/suite_test.go
```diff
@@ -44,7 +42,7 @@ var (
 			Namespace: "starboard",
 		},
 		Data: map[string]string{
-			"conftest.policy.kubernetes": `
+			"conftest.policy.runs_as_root.rego": `
```
danielpacak, line 45: *.rego suffix is required to actually eval this test policy

deven0t (approved):

Issue #514, opened 2021-04-20
Fix RBAC config when OwnerReferencesPermissionEnforcement admission controller plugin is enabled


danielpacak:
https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#ownerreferencespermissionenforcement

```yaml
# Save this to ~/tmp/kind-config/ocp-kind.config.yaml
kind: Cluster
apiVersion: kind.x-k8s.io/v1alpha4
name: ocp-kind
nodes:
- role: control-plane
  kubeadmConfigPatches:
  - |
    kind: ClusterConfiguration
    apiServer:
      extraArgs:
        enable-admission-plugins: NodeRestriction,OwnerReferencesPermissionEnforcement
- role: worker
- role: worker
```

```
kind create cluster --config ~/tmp/kind-config/ocp-kind.config.yaml
```

```
"writing report: ciskubebenchreports.aquasecurity.github.io \"ocp-kind-worker\" is forbidden: cannot set blockOwnerDeletion if an ownerReference refers to a resource you can't set finalizers on: , <nil>"
```

PR #517, opened 2021-04-21
fix: RBAC config when OwnerReferencesPermissionEnforcement admission controller plugin is enabled

danielpacak:
Resolves: #514


deploy/helm/templates/rbac.yaml
```diff
@@ -57,6 +57,12 @@ rules:
       - get
       - create
       - update
+  - apiGroups:
+      - ""
+    resources:
+      - secrets
+    verbs:
+      - delete
```
danielpacak, line 65: This is required because we set Job the owner of a secret referred to by the Job spec.

deven0t (approved):

PR #530, opened 2021-04-23
feat: Ensure default config should not create ConfigMap for Polaris when it is not chosen as plugin

danielpacak:
Resolves: #524

### DIFF
--- a/deploy/static/05-starboard-operator.config.yaml
+++ b/deploy/static/05-starboard-operator.config.yaml
@@ -211,9 +211,3 @@ kind: Secret
 metadata:
   name: starboard
   namespace: starboard-operator
----
-apiVersion: v1
-kind: ConfigMap
-metadata:
-  name: starboard-conftest-config
-  namespace: starboard-operator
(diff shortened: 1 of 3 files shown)


deploy/static/05-starboard-operator.config.yaml
```diff
@@ -211,9 +211,3 @@ kind: Secret
 metadata:
   name: starboard
   namespace: starboard-operator
----
-apiVersion: v1
-kind: ConfigMap
-metadata:
-  name: starboard-conftest-config
```
danielpacak, line 218: Static YAMLs hardcode config audit plugin to Polaris so we won't create ConfigMap for Conftest

pkg/starboard/config_test.go
```diff
@@ -324,67 +325,204 @@ func TestConfigManager_Read(t *testing.T) {
 
 func TestConfigManager_EnsureDefault(t *testing.T) {
 
-	t.Run("Should create ConfigMap with default values, and empty secret", func(t *testing.T) {
+	t.Run("Should create ConfigMaps and Secret", func(t *testing.T) {
+		g := gomega.NewGomegaWithT(t)
```
danielpacak, line 329: Gomega is more flexible and expressive than testify. We also use Gomega in integration tests so I'd like to stick to Gomega for new code.

deven0t (approved):

Issue #545, opened 2021-04-30
feat(trivy): Configure insecure image registries


danielpacak:
We do not recommend that but if someone is using the image registry with a self signed or invalid certificate, e.g. for POC, Trivy plugin will fail.
Trivy supports TRIVY_INSECURE env to skip certificate verification and pull the image after all. For a list of configured registries we should pass the env to a scan Job:

```
trivy.insecureRegistry.pocRegistry=poc.myregistry.harbor.com.pl
trivy.insecureRegistry.qaRegistry=qa.registry.aquasec.com
```

See #473

xyoxo:
Please assign me

PR #548, opened 2021-05-02
feat(trivy): Configure insecure image registries

xyoxo:
Resolves: https://github.com/aquasecurity/starboard/issues/545

### DIFF
--- a/pkg/plugin/trivy/plugin.go
+++ b/pkg/plugin/trivy/plugin.go
@@ -10,6 +10,7 @@ import (
 	"github.com/aquasecurity/starboard/pkg/kube"
 	"github.com/aquasecurity/starboard/pkg/starboard"
 	"github.com/aquasecurity/starboard/pkg/vulnerabilityreport"
+	"github.com/google/go-containerregistry/pkg/name"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -40,6 +41,7 @@ type Config interface {
 	GetTrivyImageRef() (string, error)
 	GetTrivyMode() (starboard.TrivyMode, error)
 	GetTrivyServerURL() (string, error)
+	GetTrivyInsecureRegistries() map[string]bool
 }
 
 // NewPlugin constructs a new vulnerabilityreport.Plugin, which is using an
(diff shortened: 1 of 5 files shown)


pkg/starboard/config.go
```diff
@@ -356,6 +357,17 @@ func (c ConfigData) GetTrivyServerURL() (string, error) {
 	return c.getRequiredProperty(keyTrivyServerURL)
 }
 
+func (c ConfigData) GetTrivyInsecureRegistries() []string {
```
danielpacak, line 360: I'd return a map[string]bool here so we don't duplicate the integration logic in plugin.go

danielpacak: Also we should have a unit test for new methods

pkg/plugin/trivy/plugin.go
```diff
@@ -478,6 +495,21 @@ func (s *scanner) getPodSpecForClientServerMode(spec corev1.PodSpec, credentials
 			})
 		}
 
+		ref, err := name.ParseReference(container.Image)
```
danielpacak, line 498: The code for setting the TRIVY_INSECURE env is duplicated in two places: for Standalone and ClientServer mode. Can we instead refactor this logic to a private method and invoke it twice?

pkg/plugin/trivy/plugin.go
```diff
@@ -478,6 +495,21 @@ func (s *scanner) getPodSpecForClientServerMode(spec corev1.PodSpec, credentials
...
+		if err != nil {
+			return corev1.PodSpec{}, nil, err
+		}
+		insecureRegistries := s.config.GetTrivyInsecureRegistries()
+		for _, insecureRegistry := range insecureRegistries {
+			if ref.Context().RegistryStr() == insecureRegistry {
+				env = append(env, corev1.EnvVar{
```
danielpacak, line 505: We should have a unit test scenario which asserts that the TRIVY_INSECURE env is appended.

danielpacak (changes requested):
Good start @xyoxo See my comments before we can merge this PR

pkg/starboard/config.go
xyoxo: updated to map[string]bool and added unit test.

pkg/plugin/trivy/plugin.go
xyoxo: added a private method

pkg/plugin/trivy/plugin.go
xyoxo: added unit test

danielpacak (approved):

Issue #543, opened 2021-04-29
[cli] Run Kube-bench on individual nodes


danielpacak:
See #108 

The scan ciskubebenchreports command should accept a list of node names to scan:

```
starboard scan ciskubebenchreports node-worker-1 node-worker-2 node-control-plane
```

It should fallback to scan all nodes if names where not specified:

```
starboard scan ciskubebenchreports
```

krol3:
I'd like to help with this issue.

PR #549, opened 2021-05-02
[cli] Run Kube-bench on individual nodes 

krol3:
Related to https://github.com/aquasecurity/starboard/issues/543

### DIFF
--- a/itest/starboard/starboard_cli_test.go
+++ b/itest/starboard/starboard_cli_test.go
@@ -1007,10 +1007,29 @@ var _ = Describe("Starboard CLI", func() {
 
 	Describe("Command scan ciskubebenchreports", func() {
 
+		It("scan wrong node name. It doesn't create CISKubeBenchReports", func() {
+			err := cmd.Run(versionInfo, []string{
+				"starboard",
+				"scan", "ciskubebenchreports", "notExistNode",
+				"-v", starboardCLILogLevel,
+			}, GinkgoWriter, GinkgoWriter)
+			Expect(err).ToNot(HaveOccurred())
+
+			var nodeList corev1.NodeList
+			err = kubeClient.List(context.TODO(), &nodeList)
+			Expect(err).ToNot(HaveOccurred())
+
+			for _, node := range nodeList.Items {
+				var report v1alpha1.CISKubeBenchReport
+				err := kubeClient.Get(context.TODO(), types.NamespacedName{Name: node.Name}, &report)
+				Expect(err).To(HaveOccurred(), "Not Expected CISKubeBenchReport for node %s but found", node.Name)
+			}
+		})
+
 		It("should create CISKubeBenchReports", func() {
 			err := cmd.Run(versionInfo, []string{
 				"starboard",
-				"scan", "ciskubebenchreports",
+				"scan", "ciskubebenchreports", "kind-control-plane",
 				"-v", starboardCLILogLevel,
 			}, GinkgoWriter, GinkgoWriter)
 			Expect(err).ToNot(HaveOccurred())
@@ -1095,6 +1114,7 @@ var _ = Describe("Starboard CLI", func() {
 			"cleanup",
 			"-v", starboardCLILogLevel,
 		}, GinkgoWriter, GinkgoWriter)
+		time.Sleep(4 * time.Second)
 		Expect(err).ToNot(HaveOccurred())
 	})
 
--- a/itest/starboard/starboard_cli_test.go
+++ b/itest/starboard/starboard_cli_test.go
@@ -1030,7 +1030,7 @@ var _ = Describe("Starboard CLI", func() {
 			Expect(3).To(Equal(countCISKubeBenchReport))
 		})
 
-		It("should create one CISKubeBenchReports", func() {
+		It("should create specific CISKubeBenchReports", func() {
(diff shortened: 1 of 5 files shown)


itest/starboard/starboard_cli_test.go
```diff
@@ -1102,6 +1121,7 @@ var _ = Describe("Starboard CLI", func() {
 			"cleanup",
 			"-v", starboardCLILogLevel,
 		}, GinkgoWriter, GinkgoWriter)
+		time.Sleep(4 * time.Second)
```
danielpacak, line 1124: We should use [Gomega Asynchronous Assertions](https://onsi.github.io/gomega/#making-asynchronous-assertions) instead of calling time.Sleep() explicitly

itest/starboard/starboard_cli_test.go
```diff
@@ -1014,10 +1014,29 @@ var _ = Describe("Starboard CLI", func() {
 
 	Describe("Command scan ciskubebenchreports", func() {
 
+		It("scan wrong node name. It doesn't create CISKubeBenchReports", func() {
```
danielpacak, line 1017: When you merge Describe and It it should read as the whole sentence:

*Command scan ciskubebenchreports should not create CISKubeBenchReport when node does not exist.*

```suggestion
		It("should not create CISKubeBenchReport when node does not exist", func() {
```

itest/starboard/starboard_cli_test.go
```diff
@@ -1014,10 +1014,29 @@ var _ = Describe("Starboard CLI", func() {
...
+				"scan", "ciskubebenchreports", "notExistNode",
+				"-v", starboardCLILogLevel,
+			}, GinkgoWriter, GinkgoWriter)
+			Expect(err).ToNot(HaveOccurred())
+
+			var nodeList corev1.NodeList
+			err = kubeClient.List(context.TODO(), &nodeList)
+			Expect(err).ToNot(HaveOccurred())
+
+			for _, node := range nodeList.Items {
```
danielpacak, line 1029: I think that more reliable test would be to spin up 3 nodes KIND cluster, scan just two nodes, and then assert that we only have two CISKubeBenchReport objects.

pkg/cmd/scan_ciskubebench.go
```diff
@@ -69,11 +69,20 @@ func ScanKubeBenchReports(cf *genericclioptions.ConfigFlags) func(cmd *cobra.Com
...
+			argsMap[s] = s
+		}
 
 		// TODO Move this logic to scanner.ScanAll() method. We should not mix discovery / scanning logic with the CLI.
 		var wg sync.WaitGroup
 
 		for _, node := range nodeList.Items {
+			_, found := argsMap[node.Name]
+			if !found && len(args) > 0 {
+				klog.V(3).Infof("Node not found. Skipping the scan ciskubenchreports")
```
danielpacak, line 83: If the specified node does not exist this error will be duplicated x number of nodes in the cluster. For example, on 3 nodes cluster you'll see:

```console
$ starboard scan ciskubebenchreports kind-workerr -v 3
I0504 14:22:15.008979   13121 scan_ciskubebench.go:83] Node not found. Skipping the scan ciskubenchreports
I0504 14:22:15.009172   13121 scan_ciskubebench.go:83] Node not found. Skipping the scan ciskubenchreports
I0504 14:22:15.009176   13121 scan_ciskubebench.go:83] Node not found. Skipping the scan ciskubenchreports
```

I think we should list all nodes only if the `len(args)` equals zero. Otherwise we should iterate through args and get nodes one by one.

itest/starboard/starboard_cli_test.go
```diff
@@ -1014,10 +1014,29 @@ var _ = Describe("Starboard CLI", func() {
...
+				var report v1alpha1.CISKubeBenchReport
+				err := kubeClient.Get(context.TODO(), types.NamespacedName{Name: node.Name}, &report)
+				Expect(err).To(HaveOccurred(), "Not Expected CISKubeBenchReport for node %s but found", node.Name)
```
danielpacak, line 1032: We should assert that `errors.IsNotFound(err)` is true

danielpacak (changes requested):

itest/starboard/starboard_cli_test.go
krol3: I couldn't use with success the Eventually or Consistently. For testing purpose I'm still using the sleep. I will remove it

itest/starboard/starboard_cli_test.go
krol3: I updated to spin up 3 nodes, I updated with 2 test about cisKube-bench:
- should create a specific CISKubeBenchReports: starboard scan ciskubebenchreports kind-control-plane
- should create all CISKubeBenchReport: starboard scan ciskubebenchreports

pkg/cmd/scan_ciskubebench.go
krol3: I added another control to skip the klog.V(3).Infof

danielpacak (approved):
👋 @krol3 Changes were looking good. However, I simplified a bit logic to list nodes based on input args. This will allows us to unit test `GetNodes` method instead of modifying (expensive) integration tests. I was also concerned about multi node cluster in our test env, which would take more resources.

PR #586, opened 2021-05-16
feat: Adding tolerations to scan jobs

### LINKED ISSUE #538: Add tolerations to vulnerability scan jobs spawned by Starboard
In a cluster where (all) nodes are [tainted](https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/) scan jobs created programmatically by Starboard would fail. See also #507.
One solution would be to configure tolerations at startup, e.g. in the `starboard` ConfigMap.

1. [ ] To get started add tolerations to scan Jobs scheduled by a vulnerability scanner.
2. [ ] Make sure that new code is tested

champness:
Resolves #538 

Adds tolerations to vulnerability scan jobs so that the pods can run on tainted nodes where the taints are tolerated.  Tolerations are read in JSON format from the `vulnerabilityReports.podTolerations` key of the starboard ConfigMap.


pkg/starboard/config.go
```diff
@@ -266,6 +266,8 @@ type ConfigManager interface {
 // GetDefaultConfig returns the default configuration settings.
 func GetDefaultConfig() ConfigData {
 	return map[string]string{
+		"vulnerabilityScanner.podTolerations": `[{"key":"key1","operator":"Equal","value":"value1","effect":"NoSchedule"}]`,
```
danielpacak, line 269: DITTO

pkg/vulnerabilityreport/scanner.go
```diff
@@ -115,12 +116,25 @@ func (s *Scanner) getCredentials(ctx context.Context, spec corev1.PodSpec, ns st
 	return kube.MapContainerNamesToDockerAuths(kube.GetContainerImagesFromPodSpec(spec), imagePullSecrets)
 }
 
-func (s *Scanner) prepareScanJob(workload kube.Object, spec corev1.PodSpec, credentials map[string]docker.Auth) (*batchv1.Job, []*corev1.Secret, error) {
+func getVulnerabilityScannerPodTolerations(starboardConfigData starboard.ConfigData) []corev1.Toleration {
```
danielpacak, line 119: In the future we'll have to add the same code for other types of scanners. Can we add *GetScanJobTolerations()* or a similar method to the *ConfigData* type and use it here and potentially other places?

pkg/vulnerabilityreport/scanner.go
```diff
@@ -115,12 +116,25 @@ func (s *Scanner) getCredentials(ctx context.Context, spec corev1.PodSpec, ns st
...
+	vulnerabilityScannerPodTolerations := []corev1.Toleration{}
+	tolerationsJsonFromConfigMap := starboardConfigData["vulnerabilityScanner.podTolerations"]
+	klog.V(3).Infof("vulnerabilityScanner.podTolerations: %s", tolerationsJsonFromConfigMap)
+	err := json.Unmarshal([]byte(tolerationsJsonFromConfigMap), &vulnerabilityScannerPodTolerations)
+	if err != nil {
+		klog.V(3).Infof("error unmarshalling vulnerabilityScanner.podTolerations: %s", err)
```
danielpacak, line 125: Even if the param is misconfigured (e.g. invalid JSON) would you proceed with scheduling a scan job anyway without tolerations? I think the end user won't see this Infof message unless he enabled verbose logs with the -v flag.

pkg/vulnerabilityreport/scanner_test.go
```diff
@@ -1 +1,93 @@
...
+}
+
+func (fakePlugin) ParseVulnerabilityScanResult(_ string, _ io.ReadCloser) (v1alpha1.VulnerabilityScanResult, error) {
+	return v1alpha1.VulnerabilityScanResult{}, nil
+}
+
+func TestPrepareScanJob(t *testing.T) {
+	s := &Scanner{plugin: fakePlugin{}, IDGenerator: ext.NewSimpleIDGenerator()}
+	config := starboard.ConfigData{"vulnerabilityScanner.podTolerations": `[{"key":"k"}]`}
+	job, _, err := s.prepareScanJob(kube.Object{}, corev1.PodSpec{}, map[string]docker.Auth{}, config)
```
danielpacak, line 30: Testing a private method will test only a fraction of functionality provided by *vulnerabiltyreport.Scanner* type which only exports the *Scan(context.Context, kube.Object)* method. I know that testing the *Scan* method is more challenging but that's something we should probably be doing to allow further refactoring.

Otherwise, each time you change the implementation details (private methods) we'll have to adapt unit tests. Instead of (safe) refactor we'll do code rewrite. WDYT?

pkg/vulnerabilityreport/scanner_test.go
```diff
@@ -1 +1,93 @@
...
+		expected []corev1.Toleration
+	}{
+		{
+			name:     "no vulnerabilityScanner.podTolerations in ConfigData",
+			config:   starboard.ConfigData{},
+			expected: []corev1.Toleration{},
+		},
+		{
+			name:     "vulnerabilityScanner.podTolerations value is not json",
+			config:   starboard.ConfigData{"vulnerabilityScanner.podTolerations": `lolwut`},
```
danielpacak, line 48: So how do I know when my config is invalid?

pkg/vulnerabilityreport/scanner_test.go
```diff
@@ -1 +1,93 @@
-package vulnerabilityreport_test
+package vulnerabilityreport
```
danielpacak, line 1: In other places we use a different package for test than SUT (with _test suffix) to make sure that we're only testing exported behavior.

pkg/vulnerabilityreport/scanner_test.go
```diff
@@ -1 +1,93 @@
...
+
+func TestPrepareScanJob(t *testing.T) {
+	s := &Scanner{plugin: fakePlugin{}, IDGenerator: ext.NewSimpleIDGenerator()}
+	config := starboard.ConfigData{"vulnerabilityScanner.podTolerations": `[{"key":"k"}]`}
+	job, _, err := s.prepareScanJob(kube.Object{}, corev1.PodSpec{}, map[string]docker.Auth{}, config)
+	assert.NoError(t, err)
+	assert.NotEmpty(t, job.Spec.Template.Spec.Tolerations)
+}
+
+func TestGetVulnerabilityScannerPodTolerations(t *testing.T) {
```
danielpacak, line 35: Is it a responsibility of a scanner to validate configuration? I think we should move unmarshalling / validation logic to *starboard.ConfigData* where we configure other stuff. Then we could test it along with others config-related methods.

danielpacak (changes requested):
Good start @champness I tested your changes on my end and it seems to be working fine. I left a few comments / questions to clarify the implementation details though.

deploy/helm/templates/config.yaml
```diff
@@ -6,6 +6,7 @@ metadata:
   labels:
     {{- include "starboard-operator.labels" . | nindent 4 }}
 data:
+  vulnerabilityReports.podTolerations: '[{"key":"key1", "operator":"Equal", "value":"value1", "effect":"NoSchedule"}]'
```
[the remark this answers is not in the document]
champness: Ah OK, I can move that.  The point of this default was to illustrate the usage of the config param and reflect the sample values that user is likely to have seen in the Kubernetes docs on taints and tolerations.  It seemed to me unlikely to match any real taints.  I have no objection to just removing it if you think that's cleaner.

docs/settings.md
```diff
@@ -56,27 +56,28 @@ The following tables list available configuration settings with their default va
...
+| `kube-hunter.imageRef`                | `docker.io/aquasec/kube-hunter:0.4.1`                                             | kube-hunter image reference |
+| `kube-hunter.quick`                   | `"false"`                                                                         | Whether to use kube-hunter's "quick" scanning mode (subnet 24). Set to `"true"` to enable. |
+| `polaris.imageRef`                    | `quay.io/fairwinds/polaris:3.2`                                                   | Polaris image reference |
+| `polaris.config.yaml`                 | [Check the default value here][default-polaris-config]                            | Polaris configuration file |
+| `conftest.imageRef`                   | `docker.io/openpolicyagent/conftest:v0.25.0`                                      | Conftest image reference |
+| `vulnerabilityReports.podTolerations` | `'[{"key":"key1", "operator":"Equal", "value":"value1", "effect":"NoSchedule"}]'` | JSON representation of the [tolerations](https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration) to be applied to the vulnerability scanner pods so that they can run on nodes with matching taints. |
```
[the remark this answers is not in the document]
champness: Sure that makes sense.  I can do that shortly.

pkg/cmd/scan_vulnerabilities.go
```diff
@@ -109,7 +108,7 @@ func ScanVulnerabilityReports(buildInfo starboard.BuildInfo, cf *genericclioptio
 			return err
 		}
 		scanner := vulnerabilityreport.NewScanner(kubeClientset, kubeClient, opts, plugin)
-		reports, err := scanner.Scan(ctx, workload)
+		reports, err := scanner.Scan(ctx, workload, starboardConfig)
```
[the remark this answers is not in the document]
champness: Yes, I felt that would be better too, and that was where I started but I was kinda spinning my wheels with the wiring to get it in place that way, so decided to do it this way to get the functionality completed in time.   That would be a good refactoring, I agree.

pkg/starboard/config.go
champness: I'll remove this shortly.

pkg/vulnerabilityreport/scanner.go
champness: For sure.

pkg/vulnerabilityreport/scanner_test.go
champness: That makes sense.  I can put that in the new config method.

pkg/vulnerabilityreport/scanner.go
champness: I was thinking that we would.  At least that way the user will see the pods Pending and the job incomplete and konw that they need to investigate.  Would it be better to just exit from the CLI?  What about the operator?  What should it do instead?  Still log the message but not try to schedule the job?   Perhaps I should remove the V(3)?

pkg/vulnerabilityreport/scanner_test.go
champness: I agree that testing Scan would be good and I think it would be great to do a PR just focused on that.  In this case, I felt it was beyond the scope of what I was trying to do.

I see what you're saying about testing private methods, but I don't entirely agree.  

Testing a small focused function allows you to more easily and directly exercise its capabilities across  a range of inputs.  And because it's so specific, I find that in practice it's not something that has to change often, so the concern about having to adapt tests each time you change implementation just doesn't come up all that frequently in my experience.  

In contrast, if I wanted to exercise the same 5 cases that I have in TestGetVulnerabilityScannerPodTolerations at the level of a TestScan,  I would have to do a lot of other setup that has nothing to do with the cases of interest.  And those 5 cases would be added to probably 50 or 60 other cases, interested in variations of inputs to the other private methods called by Scan.  

Instead, I'd favor specific tests for specific methods, and then the tests for Scan would only need to prove that the component methods and collaborators were wired up correctly.  In this case, TestPrepareScanJob admittedly leaves most of the details of prepareScanJob untested, but it does prove that getVulnerabilityScannerPodTolerations is wired up.

pkg/vulnerabilityreport/scanner_test.go
champness: At this point you'd need to look at logs.

pkg/vulnerabilityreport/scanner_test.go
champness: So that will rule out testing private methods, which is unfortunate in my opinion.

docs/settings.md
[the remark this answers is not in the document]
champness: Actually, as I was making this change, it occurred to me that since only vuln scans support tolerations at this point, it might be  best to keep the name of the param specific to vuln scans.  Then when the param is generalized it can be renamed to reflect that.  For now I think it would be a source of confusion to name it as if it would apply to all scans when it doesn't.  Do you still think we should rename it?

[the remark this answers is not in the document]
danielpacak: I'd rather avoid renaming config params and use a proper / generic name from the get go and apply tolerations config to other scanners before we release new version.

Deprecating config params and keeping backward compatibility is expensive. How would you migrate config objects? For example:
1. You install the Operator with vulnerabilityReport.tolerations param
2. Then we release new version of the Operator with scanJob.tolerations
3. You upgrade the Operator without modifying the starboard ConfigMap
4. You end up with invalid tolerations config (= no config) because we renamed the config key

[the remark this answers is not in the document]
champness: That's a good point.  So yeah, I'll rename the param and make note in the settings doc that it's only supported on vuln scans for the time being.

deploy/helm/templates/config.yaml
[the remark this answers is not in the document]
champness: This is done.

docs/settings.md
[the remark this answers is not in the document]
champness: This is done.

pkg/starboard/config.go
champness: This is done.

pkg/vulnerabilityreport/scanner.go
champness: This is done.

pkg/vulnerabilityreport/scanner.go
champness: With the latest refactoring, it will not schedule the scan if the tolerations are misconfigured.

pkg/vulnerabilityreport/scanner_test.go
champness: With the latest refactoring the scan will not be scheduled if the config is invalid.

pkg/vulnerabilityreport/scanner_test.go
champness: This is done.

rustrial:
Can we please also add support to pass that tolerations via Helm values?

pkg/cmd/scan_vulnerabilities.go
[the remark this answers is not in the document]
champness: That was a much better way to do it.  Thanks for the suggestion!

danielpacak (approved):
LGTM!

PR #587, opened 2021-05-17
[WIP] feat: using image digests (instead of tags) for vulnerability reports

yashvardhan-kukreja:
Signed-off-by: Yashvardhan Kukreja <yash.kukreja.98@gmail.com>

The way this PR is implemented is that it queries the pod in runtime and looks at the relevant's pod's `status.containerStatuses[*].ImageID` to identify the image digests being in use.
And starboard makes use of "ownerReferences" to identify which Pod to query as per the user-provided `starboard scan vulnerability <resource-type>/<resource-name>`.

Shedding more light to that:
- If the provided resource to be scanned is a Pod itself - For example, `starboard scan vuln.. nginx-pod` - starboard will fetch the details of that pod through controller-runtime and look into its `status.containerStatuses[*].ImageID` to identify the image digests.
- If the provided resource to be scanned is a Deployment/CronJob - For example, `starboard scan vuln...  deployment/nginx` - starboard will first of all figure out the immediate child replicaset/Job of that Deployment/CronJob respectively. And then, query only those Pods who have their ownerReference pointing to that replicaset/Job. And then, simply look at one of those Pods `status.containerStatuses[*].ImageID` (no need to look at the specs of all the pods under the replicaset/job) because all of those pods will have the same spec and hence, the same set of image digests. Hence, looking into `status.ContainerStatuses[*].ImageID` of any one child pod of the replicaset/Job would provide the relevant image digests in use.
- If the provided resource to be scanned is a ReplicaSet/DaemonSet/StatefulSet/Job - For example, `starboard scan vuln.. daemonset/sample-daemonset` - starboard will figure out the child pods of that ReplicaSet/DaemonSet/StatefulSet/Job via "ownerReference" and look at one of the fetched pods, and look at its `status.ContainerStatuses[*].ImageID` to figure out the image digests in use.

And hence, the figured out image digests will be fed to the scan job instead of image names with tags.

### DIFF
--- a/pkg/kube/object.go
+++ b/pkg/kube/object.go
@@ -154,6 +154,67 @@ func GetPodSpec(obj client.Object) (corev1.PodSpec, error) {
 	}
 }
 
+// GetPodsUnderConsideration returns []corev1.Pod from the specified
+// client.Object. Returns error if the given client.Object
+// is not a Kubernetes workload or does not point to any pod.
+func (o *ObjectResolver) GetPodsUnderConsideration(ctx context.Context, obj client.Object) ([]corev1.Pod, error) {
+	var ownerName string
+	switch t := obj.(type) {
+	case *corev1.Pod:
+		var pod corev1.Pod
+		err := o.Client.Get(ctx, types.NamespacedName{Name: obj.GetName(), Namespace: obj.GetNamespace()}, &pod)
+		if err != nil {
+			return []corev1.Pod{}, err
+		}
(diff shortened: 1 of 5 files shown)


danielpacak:
Before opening such PR we need to do a detailed impact analysis and then based on the analysis start coding. I don't know if this PR covers all corner cases because we didn't write them down. Closing for now. To resume working on it we should start with the discussion and few high level diagrams to understand what needs to be done.

PR #588, opened 2021-05-19
feat: added support for feeding custom annotations to the pods spun by scan jobs

### LINKED ISSUE #539: Add annotations to scan jobs spawned by Starboard
See #356

yashvardhan-kukreja:
closes #539 

> /kind feature

With this PR, starboard (in both cli and operator) will be able to annotate the pods, spun up by scanjobs, with user-provided custom annotations.

The custom-annotations which would be provided by the user must be in the format: `key1=value1,key2=value2,key3=value3`

* When operating in CLI mode, starboard looks into the `data.starboard.custom-annotations` field of `starboard` configmap in `starboard` namespace to get know what are user-provided custom annotations.

* When operating in Operator mode, starboard looks into the `data.starboard-operator.custom-annotations` field of `starboard` configmap in `starboard` namespace to get know what are user-provided custom annotations.

**Note**: As of now, this PR does not contain a dedicated functionality/sub-command around allowing the user input the custom-annotations in the first place. Hence, to make use of this functionality, as of now, the user would have to manually patch the starboard configmap with `data.starboard.custom-annotations` / `data.starboard-operator.custom-annotations` fields accordingly.


danielpacak (changes requested):
👋 @yashvardhan-kukreja Thank you for this PR. I left a comment to move the GetCustomAnnotationFromConfig method to ConfigData and reuse existing code.

We also prefer smaller / incremental PRs so we can remove / merge faster. For example, deal with one scanner type at a time. Otherwise changes impact all packages and in case of regression we may break everything

pkg/kube/object.go
```diff
@@ -154,6 +162,45 @@ func GetPodSpec(obj client.Object) (corev1.PodSpec, error) {
 	}
 }
 
+func (o *ObjectResolver) GetCustomAnnotationsFromConfig(ctx context.Context, executionMode ExecutionMode) (map[string]string, error) {
```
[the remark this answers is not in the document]
yashvardhan-kukreja: Sure Daniel. Will do make these suggested changes.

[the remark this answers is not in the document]
yashvardhan-kukreja: I noticed one thing that only the Scanner struct for vulnerabilityreport has the ConfigData attribute wherease rest of the Scanner structs defined under the ciskubebenchreport, configauditreport and kubehunter packages don't have ConfigData as an attribute.
So, I'll refactor all of those Scanner structs to have the ConfigData attribute so that all of them at the time of scanning can leverage the `GetScanJobAnnotations()` method so as to populate the annotations of the pods spun up by the respective scan jobs.

yashvardhan-kukreja:
> We also prefer smaller / incremental PRs so we can remove / merge faster. For example, deal with one scanner type at a time. Otherwise changes impact all packages and in case of regression we may break everything

Sure :) From next time onwards I ll ensure to segregate the PR into multiple commits (each dedicated to one sub-functionality) instead of squashing all the commits in the first place itself by creating the PR.

pkg/kube/object.go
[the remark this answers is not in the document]
danielpacak: You're right. We only injected ConfigData to vulnerability scanner

pkg/starboard/config.go
```diff
@@ -364,6 +365,35 @@ func (c ConfigData) GetTrivyMode() (TrivyMode, error) {
 		value, keyTrivyMode, Standalone, ClientServer)
 }
 
+func (c ConfigData) GetScanJobAnnotations(namespace string) (map[string]string, error) {
+	var customAnnotationsKey string
+	switch namespace {
+	case NamespaceName:
+		customAnnotationsKey = AnnotationCustomAnnotationsForScanJobPods
+	case starboardOperator.NamespaceName:
+		customAnnotationsKey = starboardOperator.AnnotationCustomAnnotationsForScanJobPods
```
danielpacak, line 374: Define a new param called `scanJob.annotations`. Similar to what we did for `scanJob.tolerations`. There's no need to prefix a key for CLI and operator. They have their own copy of the ConfigMap. So just read key from the map

pkg/operator/etc/constants.go
```diff
@@ -0,0 +1,23 @@
+package etc
+
+const (
+	// NamespaceName the name of the namespace in which Starboard-operator stores its
+	// configuration and where it runs scan jobs.
+	NamespaceName = "starboard-operator"
+
+	// ServiceAccountName the name of the service account used to provide
+	// identity for scan jobs run by Starboard-operator.
+	ServiceAccountName = "starboard-operator"
```
danielpacak, line 10: Service account is configurable through ENV. Also these constants are duplication what we already have in the starboard package. Why do you need them?

yashvardhan-kukreja:
Ohh I pushed by mistake haha. Still have to clean some bits. Meanwhile, I ll address your comments as well :)

pkg/starboard/config.go
```diff
@@ -457,7 +487,9 @@ func (c *configManager) EnsureDefault(ctx context.Context) error {
 		if !apierrors.IsNotFound(err) {
 			return fmt.Errorf("getting config: %w", err)
 		}
-
+		configData := GetDefaultConfig()
```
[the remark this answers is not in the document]
yashvardhan-kukreja: Yes, actually, I was about to get rid of this but before that I mistakenly pushed my code and it got fed here 😛

danielpacak:
> @danielpacak I noticed all your comments. Most of them are basically, a root cause of the fact that I tried. to keep different key names in configmaps in the case of CLI mode (`starboard.custom-annotations`) and Operator (`starboard-operator.custom-annotations`) mode.
> So, as suggested, I should get rid of this difference and just decide on a common key name like `starboard.custom-annotations` in both CLI and Operator mode.
> That would reduce a good chunk of redundant code and maintain the namespace agnosticism with respect to `ConfigData`.

As I mentioned please use naming convention consistent with tolerations, i.e. `scanJob.annotations`. We already have param named `scanJob.tolerations` and we'll add `scanJob.resourceRequests`, etc. The reason for that is that we plan to change to structured Starboard configuration object because key-value pairs are quite limited and cumbersome to parse. Anyway, I expect that in this PR we'll have sth like:

```yaml
data:
  scanJob.tolerations: "[{"key": "key1", "value": "value1", "operator": "Equal"}]"
  scanJob.annotations: "key1=value1,key2=value2"
```

It's not in scope of this issue / PR but we'll change one day to:

```yaml
config:
  scanJob:
    tolerations:
      - key: key1
        value: value1
        operator: Equal
    annotations:
      my.annotation1: "some value"
      my.annotation2: "some other value"
```

yashvardhan-kukreja:
Oh yes yes, I was just exemplifying. In reality, I am going to do with the respect to the way you suggested here i.e. `scanJob.annotations`
> Define a new param called scanJob.annotations. Similar to what we did for scanJob.tolerations. There's no need to prefix a key for CLI and operator. They have their own copy of the ConfigMap. So just read key from the map

yashvardhan-kukreja:
Quick overview of the current functionality:

I fed this configmap:

```yaml
apiVersion: v1
data:
  configAuditReports.scanner: Polaris
  kube-bench.imageRef: docker.io/aquasec/kube-bench:0.5.0
  polaris.imageRef: quay.io/fairwinds/polaris:3.2
  scanJob.annotations: foo=bar,Creedthoughts.Gov.Www/creedthoughts=creed.blog. ### the scan job annotations
  trivy.httpProxy: ""
  trivy.httpsProxy: ""
  trivy.imageRef: docker.io/aquasec/trivy:0.16.0
  trivy.mode: Standalone
  trivy.noProxy: ""
  trivy.severity: UNKNOWN,LOW,MEDIUM,HIGH,CRITICAL
  vulnerabilityReports.scanner: Trivy
kind: ConfigMap
metadata:
  name: starboard
  namespace: starboard-operator
```

And that's how a pod spun by vulnerability reports looked

```yaml
apiVersion: v1
kind: Pod
metadata:
  annotations:
    foo: buddysup
    Creedthoughts.Gov.Www/creedthoughts: creed.blog
  creationTimestamp: "2021-05-20T21:11:30Z"
  generateName: scan-vulnerabilityreport-59c955884-
  labels:
    app.kubernetes.io/managed-by: starboard
    controller-uid: c034578e-77be-4351-b8e8-a8bf5a4c8ca3
    job-name: scan-vulnerabilityreport-59c955884
    pod-spec-hash: 694c58c54
    starboard.resource.kind: Pod
    starboard.resource.name: nginx
    starboard.resource.namespace: default
    vulnerabilityReport.scanner: "true"
  name: scan-vulnerabilityreport-59c955884-rxnxx
  namespace: starboard-operator
  ownerReferences:
  - apiVersion: batch/v1
    blockOwnerDeletion: true
```

Same thing happened for the pods spun by configauditreports, etc.
Same for CLI mode as well on execution `starboard scan ...`

yashvardhan-kukreja:
Noice, looking at the green flags, I ll make the relevant changes to `settings.md` as well

pkg/kubebench/scanner.go
```diff
@@ -20,28 +20,34 @@ import (
...
+	logsReader      kube.LogsReader
+	plugin          Plugin
+	objectResolver  *kube.ObjectResolver
+	starboardConfig starboard.ConfigData
 }
 
 func NewScanner(
-	scheme *runtime.Scheme,
 	clientset kubernetes.Interface,
+	client client.Client,
```
danielpacak, line 38: We don't need client.Client so long we remove injected ObjectResolver

pkg/kubehunter/scanner.go
```diff
@@ -34,23 +35,25 @@ type Scanner struct {
 	config    Config
 	clientset kubernetes.Interface
 	ext.IDGenerator
-	opts       kube.ScannerOpts
-	logsReader kube.LogsReader
+	opts           kube.ScannerOpts
+	logsReader     kube.LogsReader
+	objectResolver *kube.ObjectResolver
```
danielpacak, line 40: We are not using ObjectResolver

pkg/kubehunter/scanner.go
```diff
@@ -34,23 +35,25 @@ type Scanner struct {
...
 }
 
 func NewScanner(
-	scheme *runtime.Scheme,
-	config Config,
 	clientset kubernetes.Interface,
+	client client.Client,
```
danielpacak, line 45: We are not using client.Client

pkg/kubehunter/scanner.go
```diff
@@ -109,6 +112,11 @@ func (s *Scanner) prepareKubeHunterJob() (*batchv1.Job, error) {
 		kubeHunterArgs = append(kubeHunterArgs, "--quick")
 	}
 
+	scanJobAnnotations, err := s.config.(starboard.ConfigData).GetScanJobAnnotations()
```
danielpacak, line 115: We should inject ConfigData and avoid casting to wider interface

pkg/starboard/config_test.go
```diff
@@ -704,3 +704,53 @@ CVE-2019-1543`,
...
+	}
+
+	for _, tc := range testcases {
+		scanJobAnnotations, err := tc.config.GetScanJobAnnotations()
+		if tc.expectError != "" {
+			assert.Error(t, err, tc.expectError, tc.name)
+		} else {
+			assert.NoError(t, err, tc.name)
+		}
+		assert.Equal(t, tc.expected, scanJobAnnotations, tc.name)
```
danielpacak, line 754: This should go under the else scope?

danielpacak (changes requested):
Great progress @yashvardhan-kukreja Hopefully one last round and we'll merge it!

pkg/starboard/config_test.go
yashvardhan-kukreja: So, in all the cases, even including errors, a scanJonAnnotations object is generated just that in case of errors, it is `map[string]string{}` i.e. an empty map. So, I included that as well as a part of the tests.

pkg/kubebench/scanner.go
```diff
@@ -20,28 +20,34 @@ import (
...
-	opts       kube.ScannerOpts
-	clientset  kubernetes.Interface
-	logsReader kube.LogsReader
-	plugin     Plugin
+	scheme          *runtime.Scheme
+	opts            kube.ScannerOpts
+	clientset       kubernetes.Interface
+	logsReader      kube.LogsReader
+	plugin          Plugin
+	objectResolver  *kube.ObjectResolver
```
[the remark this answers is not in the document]
yashvardhan-kukreja: Oh yes! Forget to revert this. Will do so

pkg/starboard/config_test.go
danielpacak: Ah got it!

pkg/kubehunter/scanner.go
yashvardhan-kukreja: Daniel, I guess to tackle this what we can do is that from the struct here, we can remove the `config Config` and add `starboardConfig starboard.ConfigData` in place of it instead.

So, this wouldn't require refactoring any of the existing functionality as `Config` type is an interface and `starboard.ConfigData` struct has the relevant methods to make it implement that interface. Meaning that previously whatever methods were being called upon `config Config` (in the above scanner struct of kubehunter) can be, without any change, called upon by `starboardConfig starboard.ConfigData`. In this case, `GetKubeHunterImageRef()` is the method.

PS: I am suggesting this instead of straight away adding `starboardConfig starboard.configData` without getting rid of `config Config`, because doing so would add unnecessary redundancy as `config Config` wouldn't be serving any special purpose proving its necessity/need in the struct as whatever it is doing can be simply done by `starboardConfig starboard.ConfigData` itself.

I hope that makes sense.

danielpacak: Okey. Let's leave it like it is for now. We can refactor it later on.

danielpacak (approved):

PR #616, opened 2021-06-04
refactor(trivy): Read config from starboard-trivy-config ConfigMap

danielpacak:
We started off with all config settings read from the starboard
ConfigMap and the starboard Secret. However, this does not scale
well with a number of plugins and configuration options.

This commit allows Triy plugin to read its settings from
the starboard-trivy-config ConfigMap and the starboard-trivy-config
Secret.


itaysk:
I skimmed through it. looks good, but I wasn't sure if you needed to also update the config defaults for the init command use-case?

danielpacak:
> I skimmed through it. looks good, but I wasn't sure if you needed to also update the config defaults for the init command use-case?

That's a good point. I've updated the PR to ensure default plugin config when init command is run. Previously I deferred it to scan commands. It was easier in terms of code, but not very intuitive. Now all ConfigMaps are created in the init command.

itaysk (approved):

PR #618, opened 2021-06-08
chore: chart app version

krol3:
- Upgrade the helm chart with the release tag app.
- The helm chart will be the same version as the app.
 `helm package --app-version=${RELEASE} --version=${RELEASE}`

### DIFF
--- a/.github/workflows/publish-helm-chart.yaml
+++ b/.github/workflows/publish-helm-chart.yaml
@@ -4,6 +4,8 @@
 name: Publish Helm chart
 
 on:
+  tags:
+    - "v*"
   workflow_dispatch:
 env:
   HELM_REP: helm-charts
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-helm-chart.yaml
```diff
@@ -20,14 +22,17 @@ jobs:
...
         run: |
-          wget https://github.com/helm/chart-releaser/releases/download/v1.1.1/chart-releaser_1.1.1_linux_amd64.tar.gz
-          tar xzvf chart-releaser_1.1.1_linux_amd64.tar.gz cr
+          wget "https://github.com/helm/chart-releaser/releases/download/v${VERSION}/chart-releaser_${VERSION}_linux_amd64.tar.gz"
+          tar xzvf chart-releaser_${VERSION}_linux_amd64.tar.gz cr
       - name: Package helm chart
         run: |
-          ./cr package ${{ env.CHART_DIR }}
+          RELEASE=$(echo ${{ github.ref }} | sed -e "s#refs/tags/##g" | sed -e 's/^v//')
+          helm package --app-version=${RELEASE} --version=${RELEASE} ${{ env.CHART_DIR }} -d .cr-release-packages
```
danielpacak, line 35: If we take version from git revision how and when will we update versions in Chart.yaml https://github.com/aquasecurity/starboard/blob/main/deploy/helm/Chart.yaml#L9 and https://github.com/aquasecurity/starboard/blob/main/deploy/helm/Chart.yaml#L14

For now, before each release I update Chart.yaml and then trigger this build so even if it's kinda manual the value is src code is consistent with the version of published chart.

.github/workflows/publish-helm-chart.yaml
```diff
@@ -4,6 +4,8 @@
 name: Publish Helm chart
 
 on:
+  tags:
+    - "v*"
```
danielpacak, line 8: So far it was manually dispatched workflow. If we want to publish on each release, it should probably a step executed in the Release workflow after we publish docker images. WDYT?

.github/workflows/publish-helm-chart.yaml
krol3: In this PR, the chart version and the chart app version will be the same. I will use the simple 1-1 versioning. I tried to follow this approach: https://codefresh.io/docs/docs/new-helm/helm-best-practices/

danielpacak (changes requested):
I'm not sure that merging the chart version and app version is a good idea. There are some cases where we can benefit from both. Therefore, lets stick to existing versioning and Git as single source of truth when it comes to tagging releases and Helm charts.

We can restart discussion on then when we have a real problem with current versioning.

PR #622, opened 2021-06-14
feat: Define ClusterConfigAuditReport

danielpacak:
We could not associate ConfigAuditReports with cluster-scoped
resources such as ClusterRole or CustomResourceDefinition.
Therefore, this commit defines cluster-scoped equivalent to
ConfigAuditReport called ClusterConfigAuditReport.


deven0t (approved):
LGTM

PR #650, opened 2021-08-02
chore: udpate report image documentation

krol3:
Add an image sample about the report generated.


docs/cli/getting-started.md
```diff
@@ -26,8 +26,10 @@ starboard init
 The `init` subcommand creates the `starboard` namespace, in which Starboard executes Kubernetes jobs to perform
 scans. It also sends custom security resources definitions to the Kubernetes API:
 
-```console
+```
```
danielpacak, line 29: Was it intentional? To split command and output?

krol3: Yes! We could leave only the text that It's relevant to copy to the clipboard. And not copy all the output too.

danielpacak (approved):

Issue #631, opened 2021-06-26
Upgrade kube-hunter to v0.6.1


danielpacak:
See https://github.com/aquasecurity/kube-hunter/issues/465

PR #690, opened 2021-08-29
FIX: Updated log level as temporary solution

py-go:
This PR includes a temporary fix for **log** related problem.
Issue ID: https://github.com/aquasecurity/starboard/issues/631
Solution Reference: https://github.com/aquasecurity/kube-hunter/issues/465


danielpacak (approved):
LGTM

PR #711, opened 2021-09-18
Added intial version of image digest scanner

py-go:


danielpacak:
Closing this PR as is for now. We must design reliable solution and consider all corner cases before we proceed with Go code.

PR #725, opened 2021-10-01
chore(ci): Add KinD connection test to `integration-operator-conftest`

### LINKED ISSUE #652: chore(ci): Fail fast by checking if K8s KIND cluster is ready before wer run integration tests

czunker:
Fixes #652

Signed-off-by: Christian Zunker <christian.zunker@codecentric.cloud>

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -7,22 +7,22 @@ on:
     branches:
       - main
     paths-ignore:
-      - '.github/ISSUE_TEMPLATE/*.md'
-      - '*.md'
-      - 'docs/**'
-      - 'mkdocs.yml'
-      - 'LICENSE'
-      - 'NOTICE'
+      - ".github/ISSUE_TEMPLATE/*.md"
+      - "*.md"
+      - "docs/**"
+      - "mkdocs.yml"
+      - "LICENSE"
+      - "NOTICE"
   pull_request:
     branches:
       - main
     paths-ignore:
-      - '.github/ISSUE_TEMPLATE/*.md'
-      - '*.md'
-      - 'docs/**'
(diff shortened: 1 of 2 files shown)


czunker:
For the other integration tests, it was already implemented:
https://github.com/aquasecurity/starboard/blob/main/.github/workflows/build.yaml#L101
https://github.com/aquasecurity/starboard/blob/main/.github/workflows/build.yaml#L139
and
https://github.com/aquasecurity/starboard/blob/main/.github/workflows/release.yaml#L54
https://github.com/aquasecurity/starboard/blob/main/.github/workflows/release.yaml#L87

.github/workflows/build.yaml
```diff
@@ -184,6 +184,10 @@ jobs:
         with:
           version: ${{ env.KIND_VERSION }}
           image: ${{ env.KIND_IMAGE }}
+      - name: Test connection to Kubernetes cluster
+        run: |
+          kubectl cluster-info
+          kubectl describe node
```
danielpacak, line 190: AFAIK this will only describe node even if it's is `NotReady` status. Ideally kind action should block and wait until nodes are ready, but what I've seen it's not the case.

Therefore, I believe we should `kubectl wait` for node before we run integration tests.

danielpacak (changes requested):

.github/workflows/build.yaml
czunker: Ah, learned something new. Thanks.

I added the requested wait. Do you still want to keep the `describe` for the information?

danielpacak: 🚀 Description might be useful one day.

danielpacak (approved):

PR #733, opened 2021-10-05
feat(crd): Remove the all category from cluster-scoped CRDs

danielpacak:
This patch will not display cluster-scoped security reports
when kubectl get all command is run for the specified namespace.

    kubectl get all -n kube-system


yoavrotems (approved):
LGTM

PR #750, opened 2021-10-15
chore: documentation getting started

krol3:
Improving the copy and paste console, and using results table for the results.


krol3:
Hi @danielpacak , I tried to follow the same approach about trivy documentation. 
Using in the markdown the details and summary. Any improvement or comment about this? 

<details>
<summary>Result</summary>

danielpacak:
> Hi @danielpacak , I tried to follow the same approach about trivy documentation. Using in the markdown the details and summary. Any improvement or comment about this?
> 
> Result

I know the `<details>` and `<summary>` tags. But personally, I don't really like it because when you print the documentation page (e.g. to PDF) the layout and highlight is broken. For the the MkDocs should fix the copy experience when you use console as code block language.

<pre>
```console
$ only the line prefixed with the dollar sign should be copied
output of the command
output of the command
```
</pre>
```console
$ name -a
Darwin dpacak-imac.local 21.2.0 Darwin Kernel Version 21.2.0: Sun Nov 28 20:28:54 PST 2021; root:xnu-8019.61.5~1/RELEASE_X86_64 x86_64
```

 Anyway, I let @chen-keinan decide if we want to use it.

chen-keinan:
I agree that this PR in a way is kind of a workaround to the problem; however I feel we should go with it because:
- It solves the problem - shell commands only copied
- It makes the docs look clean
- Its look and feel is similar to Trivy (which is important for our open source docs to look alike)

docs/cli/getting-started.md
```diff
@@ -97,6 +104,22 @@ Retrieve the configuration audit report:
 starboard get configauditreports deployment/nginx -o yaml
 ```
 
+or
+
+```console
+kubectl get configauditreport -o wide \
```
danielpacak, line 110: We no longer recommend using kubectl get configauditrepors as the naming convention and label selectors might change. In other works labels set is an implementation details. Also for some K8s object, that contain semicolons we cannot use labels and we fallback to annotations. See the code for details.

danielpacak (changes requested):
Before we merge this one please see my comment about labels and implementation details. Otherwise I'm okey with details and summary tags.

danielpacak (approved):

Issue #290, opened 2020-12-23
Cannot use SecurityContext


djkormo:
How to configure starboard via operator to use securityContext ?

k get pod,job -n starboard-operator
NAME                                             READY   STATUS                            RESTARTS   AGE
pod/16f0140e-2769-4b89-9d0e-c56c9f317bc2-4n4xm   0/1     Init:CreateContainerConfigError   0          33s

pod/starboard-operator-6759cbc887-2zzjz          1/1     Running                           0          5m36s

NAME                                             COMPLETIONS   DURATION   AGE
job.batch/16f0140e-2769-4b89-9d0e-c56c9f317bc2   0/1           33s        33s

**Error: container has runAsNonRoot and image will run as root**

**Environment:**

- Starboard version starboard-operator:0.7.1
- Kubernetes version 1.18.9
- OS (Windows 10):

danielpacak:
👋 @djkormo Some scanners must run as root, e.g. Trivy v0.14.0 - the default vulnerability scanner integrated with Starboard Operator. Hence, we do not allow configuring SecurityContext for pods run by vulnerability scan jobs. Polaris, on the other hand, which is integrated with Starboard CLI, can run as non root.

We still have some tasks to implement least-privileged principle for other scanners, but in general SecurityContext won't be configurable, because invalid config might actually break the functionality.

djkormo:
So maybe  you shoud prepare manifests with  rbac for service account ?

danielpacak:
Well, we have RBAC rules for service account that is running the operator's pod and downstream scan jobs. See, for example, https://github.com/aquasecurity/starboard/blob/main/deploy/helm/templates/rbac.yaml
I don't see how is that related to SecurityContext though?

djkormo:
Maybe I'm wrong, but I cannot see  permission like this

apiVersion: rbac.authorization.k8s.io/v1
kind: Role
metadata:
  name: starboard-operator-psp
  namespace: starboard-operator
rules:
- apiGroups:
  - policy
  resourceNames:
  - privileged
  resources:
  - podsecuritypolicies
  verbs:
  - use

danielpacak:
Okey so maybe without jumping into solutions straight away we should better understand what is the problem that you're trying to solve? I guess this is related to PodSecurityPolices enabled in your cluster. Could you provide more details about this setup? In particular what are the constraints that you usually put on workloads running in the cluster. Based on that we can think of RBAC config and / or SecurityContexts

[...]

PR #771, opened 2021-10-25
feat(helm): Add PodSecurityPolicy creation

aarnq:
Add option to create a PodSecurityPolicy to allow the Trivy scanner to run within a cluster using PSPs. 
It is the minimal possible to allow the scanner pods to run as root, and fix the "*container has runAsNonRoot and image will run as root*" error when the default PSP restricts it.
Related to #290. 

I've set the name to start with `00` as there seems to be nothing set in the `securityContext` of the Trivy scanners to actually specify that they will run as root. This makes them use the alphabetically first PSP that allows the Pod spec to run, which in our case is the default restrictive one that allows the Pod but not the image.

I don't know where or how to change the spec of the Trivy scanners created by the operator so I'm focusing on what I used to make it run.

### DIFF
--- a/deploy/helm/templates/psp.yaml
+++ b/deploy/helm/templates/psp.yaml
@@ -1,0 +1,39 @@
+{{- if .Values.podSecurityPolicy.create -}}
+---
+apiVersion: policy/v1beta1
+kind: PodSecurityPolicy
+metadata:
+  name: {{ include "starboard-operator.fullname" . }}
+  labels:
+    {{- include "starboard-operator.labels" . | nindent 4 }}
+spec:
+  allowPrivilegeEscalation: false
+  fsGroup:
+    ranges:
+      - max: 65535
+        min: 1
+    rule: MustRunAs
+  requiredDropCapabilities:
+    - ALL
+  runAsGroup:
+    ranges:
+      - max: 65535
+        min: 1
+    rule: MustRunAs
+  runAsUser:
+    rule: RunAsAny
+  seLinux:
+    rule: RunAsAny
+  supplementalGroups:
+    ranges:
+      - max: 65535
+        min: 1
+    rule: MustRunAs
+  volumes:
+    - configMap
+    - emptyDir
+    - projected
+    - secret
+    - downwardAPI
+    - persistentVolumeClaim
+{{- end}}
--- a/deploy/helm/templates/rbac.yaml
+++ b/deploy/helm/templates/rbac.yaml
@@ -139,6 +139,16 @@ rules:
       - get
       - update
   {{- end }}
+  {{- if .Values.podSecurityPolicy.create }}
+  - apiGroups:
+      - extensions
+    resourceNames:
+      - {{ include "starboard-operator.fullname" . }}
+    resources:
+      - podsecuritypolicies
+    verbs:
+      - use
+  {{- end}}
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: {{ $conditionalClusterPrefix }}RoleBinding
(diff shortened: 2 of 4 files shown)


aarnq:
Sorry, I've been neglecting this PR for quite some time. I'll take a new look at it and update it.

aarnq:
With the default setup (no PSP) the Trivy container has the following SecurityContext:
```yaml
securityContext:
  allowPrivilegeEscalation: false
  capabilities:
    drop:
      - ALL
      - all
  privileged: false
  readOnlyRootFilesystem: true
  runAsGroup: 1
  runAsNonRoot: true
```

With a PSP called `starboard-operator` allowed by the `starboard-operator` ServiceAccount generated by the helm chart the `starboard-operator*` Pod got the PSP, however the `scan-vulnerabilityreport*` Pods didn't.

With a PSP called `00-starboard-operator` both the `starboard-operator*` and `scan-vulnerabilityreport*` Pods got the PSP, and this gave the Trivy container the following SecurityContext:
```yaml
securityContext:
  allowPrivilegeEscalation: false
  capabilities:
    drop:
      - ALL
      - all
  privileged: false
  readOnlyRootFilesystem: true
```

Does starboard-operator actually specify a SecurityContext for the Trivy scanner that it will run as root or is it just set in the image?

aarnq:
I manged to add `runAsNonRoot: false` to the SecurityContext of the Trivy container in `pkg/plugin/trivy/plugin.go`.
This does allow it to properly use the `starboard-operator` PSP allowing it to run as root.

Do you want me add it here or should I open a new PR for it?
Or do you want me to open an issue about this whole thing since this got a bit more involved then I initially thought. :sweat_smile:

danielpacak:
> 

Thank you for digging into this @aarnq I think we can do all the work related to Trivy and PSP in scope of this PR. I just wanted to better understand the reproduction steps when Trivy scan job fails after you enable PSP admission controller. Could you share the default policy that blocks vulnerability scan jobs so we can see exact errors?

aarnq:
Sure, I can do that @danielpacak. This is when running on version `v0.13.1`.

For the default "restricted" PSP we have this defined:
```yaml
spec:
  allowPrivilegeEscalation: false
  fsGroup:
    ranges:
    - max: 65535
      min: 1
    rule: MustRunAs
  requiredDropCapabilities:
  - ALL
  runAsGroup:
    ranges:
    - max: 65535
      min: 1
    rule: MustRunAs
  runAsUser:
    rule: MustRunAsNonRoot
  seLinux:
    rule: RunAsAny
  supplementalGroups:
    ranges:
    - max: 65535
      min: 1
    rule: MustRunAs
  volumes:
  - configMap
  - emptyDir
  - projected
  - secret
  - downwardAPI
  - persistentVolumeClaim
```
Then the scan-vulnerabilityreports would be in the state `CreateContainerConfigError` with the message :
```yaml
message: 'container has runAsNonRoot and image will run as root (pod: "scan-vulnerabilityreport-[...]",
          container: [...])'
```
And they get the `securityContext` listed first in the comment above.

Let know if you want more.

danielpacak:
👋 @aarnq I'm sorry it took us so long to finally look into your PR. I was actually wondering if it still makes sense to maintain a deprecated PodSecurityPolicy resource that eventually will be removed as explained in the [PodSecurityPolicy Deprecation: Past, Present, and Future](https://kubernetes.io/blog/2021/04/06/podsecuritypolicy-deprecation-past-present-and-future/) blog.

How does it sound if we just document how to use Starboard in a cluster that's still using deprecated [PodSecurityPolicy](https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#podsecuritypolicy) admission controller?

aarnq:
No worries @danielpacak. Yeah sure, that sounds fine. :+1: I'll close this.
Thanks.

PR #819, opened 2021-11-21
chore: helm testing gh-action

krol3:
Adding a lint and testing step in github action to the helm chart

### DIFF
--- a/.github/workflows/publish-helm-chart.yaml
+++ b/.github/workflows/publish-helm-chart.yaml
@@ -9,6 +9,8 @@ env:
   HELM_REP: helm-charts
   GH_OWNER: aquasecurity
   CHART_DIR: deploy/helm
+  KIND_VERSION: "v0.11.1"
+  KIND_IMAGE: "kindest/node:v1.21.1@sha256:69860bda5563ac81e3c0057d654b5253219618a22ec3a346306239bba8cfa1a6"
 jobs:
   release:
     runs-on: ubuntu-20.04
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-helm-chart.yaml
```diff
@@ -21,6 +23,20 @@ jobs:
...
           version: v3.5.0
+      - name: Set up python
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.7
+      - name: Setup Chart Linting
+        id: lint
+        uses: helm/chart-testing-action@v2.1.0
+      - name: Setup Kubernetes cluster (KIND)
+        uses: engineerd/setup-kind@v0.5.0
```
danielpacak, line 34: Why don't we use https://github.com/helm/kind-action instead of engineerd/setup-kind?

.github/workflows/publish-helm-chart.yaml
```diff
@@ -21,6 +23,20 @@ jobs:
         uses: azure/setup-helm@v1
         with:
           version: v3.5.0
+      - name: Set up python
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.7
+      - name: Setup Chart Linting
+        id: lint
+        uses: helm/chart-testing-action@v2.1.0
```
danielpacak, line 32: General comments for this and other workflow definition files:
1. Given all these supply chain attack I'd rather pin a version of GitHub action to a Git commit or Docker registry digest instead of (potentially mutable) tags.
2. We should check integrity of files downloaded with wget | curl or switch to Docker actions (e.g. check integrity of chart-releaser_1.1.1_linux_amd64.tar.gz)

krol3: I considered your comments, please take a look, when you have time @danielpacak

danielpacak (approved):
LGTM!

PR #829, opened 2021-11-24
chore: test the helm chart in the PR

krol3:
- Run the testing helm chart in the PR (only 'deploy/helm/**')
- Separate the release chart in another job.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -24,6 +24,7 @@ on:
       - 'mkdocs.yml'
       - 'LICENSE'
       - 'NOTICE'
+
 env:
   GO_VERSION: "1.16"
   KIND_VERSION: "v0.11.1"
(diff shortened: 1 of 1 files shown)


krol3:
@danielpacak please validate the idea of test the manifest of the helm in the PR, if make sense.

danielpacak (changes requested):
Thank you for working on this @krol3 ! I think it's great idea to test the Helm chart as part of our build pipeline.

We have already grouped all checks related to pull requests in a single file called `.github/worflow/build.yaml` so I'd add such job in there and run it after all integration tests passed for few reasons:
- If tests fail we gonna restart build anyway so doing all these Helm checks would be wasteful.
  starboard-ci-workflow
- Maybe not in scope of this PR but ... after creating a Helm release it would be very nice to run actual integration tests to see the the operator deployed with Helm really works (not just that Helm release was created).
- We'd like to get rid of the `workflow_dispatch` event and publish to Helm repository (and MkDocs) when the release workflow is complete.

krol3:
@danielpacak the "Test Helm Chart" was moved to the build job.

danielpacak (approved):
Thanks for working on this @krol3 I just slightly updated the workflow to run Helm chart tests only after integration tests passed (as originally suggested) 🍀 🌏 .

Issue #818, opened 2021-11-19
Write proposal to use Trivy file system scanner


danielpacak:
See the initial POC described in #692

Additional requirements (/cc @jerbia):
1. Whether to run the scanning pod on the same host (or not) should be configurable
2.  If scanning pod is not scheduled on the same host, then the image pull policy should allow pulling the image

PR #830, opened 2021-11-26
docs(design): Scan Container Images with Trivy Filesystem Scanne

deven0t:
resolves: #818


docs/design/design_trivy_file_system_scanner.md
```diff
@@ -0,0 +1,200 @@
...
+Currently, clientServer mode is not supporting filesystem scan. So we will have to go with standalone mode. In future we
+ can change it to client server mode and accordingly we will have to change our scan job spec.
+ 
+ Configuration parameter can be added to enable this scan mode
+``trivy.command: fs``. This trivy.command will support two values ``image`` and ``fs``. For backward compatibility. if 
+this config parameter is not defined then it will consider it as ``image``.
+
+We have two options to consider here
+1. We can restrict scan job to run on same host where original pod is running.
+    - ImagePullPolicy can be set as false, so we will not need to pull the image on node
```
danielpacak, line 52:
```suggestion
    - ImagePullPolicy will be set to `Never`, so we will not need to pull the image on node
```

docs/design/design_trivy_file_system_scanner.md
```diff
@@ -0,0 +1,200 @@
...
+
+To run scan on same host where application pod is running then we have to identify host on which application pod is 
+running. To do this we will search for active pod for any controller and we will get the nodename from first pod of it.
+
+1. Scan job has to run on same host where pod is running.
+    - Get the nodename of one of the pod. If we dont get nodename then return error.
+    - No need to copy secret from the source namespace to starboard-operator namespace. We will ignore imagePullSecret of 
+    source application workload.
+    - ImagePullPolicy will be set as Never
+    - nodeName will be passed explicitly.
```
danielpacak, line 69: Clarify that we set nodeName on a scan job so it's scheduled on the same node as scanned workload.

docs/design/design_trivy_file_system_scanner.md
```diff
@@ -0,0 +1,200 @@
...
+            - --ignore-unfixed
+            - /
+          volumeMounts:
+            - name: scan-volume
+              mountPath: /var/starboard
+```
+
+##### Current Limitations for this option:
+1. If controller is created, and it doesn’t have any active pods then this mode will not scan it
+2. We cannot identify pods running for cronjob, so those will not be scanned.
```
danielpacak, line 194: Can you elaborate on why we cannot identify pods running for CronJobs? Is it because we scan by K8s controller not by pod?

docs/design/design_trivy_file_system_scanner.md
```diff
@@ -0,0 +1,200 @@
...
+        # The scan-nginx container is based on the container image that
+        # we want to scan with Trivy. However, it has overwritten command (entrypoint)
+        # to invoke trivy file system scan. The scan results are output to stdout
+        # in JSON format so we can parse them and store as VulnerabiltyReport instance.
+        - name: scan-nginx
+          image: example.registry.com/nginx:1.16
+          # To scan image layers cached on K8s node without pulling
+          # it from a remote registry.
+          imagePullPolicy: Never
+          command:
```
danielpacak, line 178: We are changing the entrypoint here, but I think we should also mention that Trivy requires root permissions. IIRC a scan Job must run as UID 0. Does it have any impact on PSP or any solution that supersedes PSPs?

docs/design/design_trivy_file_system_scanner.md
```diff
@@ -0,0 +1,200 @@
...
+With this approach we have following problem:
+1. When image pulled from private registry without ImagePullSecret or service account 
+(https://kubernetes.io/docs/concepts/containers/images/#configuring-nodes-to-authenticate-to-a-private-registry)  
+2. When image pulled from managed registry of managed cluster
+3. When pod is running with image present on host
+
+## Proposal
+
+Use Trivy file system scanning option to scan an image. So that we will not need to provide registry credentials to trivy.
+K8s scan job will use underline secrets available to pull an image and that image will be scanned by trivy as file system
```
danielpacak, line 24: For the current proposal the main idea is to schedule a scan job on the same node where the scanned workload so we can leverage container image cached on that node and scan without providing credentials to Trivy.

danielpacak (changes requested):
Hey @deven0t Overall LGTM. I left a few comments to clarify the proposal.

docs/design/design_trivy_file_system_scanner.md
deven0t: yes, boz when starboard receives event of cronjob we may not have an active pod running for it. and also getting active pods of cron job is not straight forward way, so added it as limitation for now.
this is one way to get jobs for cronjob, later we will have to get active pods for these jobs. https://github.com/kubernetes/kubernetes/blob/e640a01219736dd051dc3a9b8ac4630ec18b6b25/pkg/controller/cronjob/cronjob_controllerv2.go#L226

docs/design/design_trivy_file_system_scanner.md
deven0t: Yes, we need to run this scan job container with root user `runAsUser: 0`. This will have impact with PSP MustRunAsNonRoot

danielpacak (approved):
Great work @deven0t I just pushed a few edits to your branch to cleanup duplicates and some of the implementations details.

docs/design/design_trivy_file_system_scanner.md
```diff
@@ -0,0 +1,197 @@
...
+the container image which is already cached on that node without pulling the image from a remote registry. What's more,
+Trivy will scan container images from private registries without providing registry credentials (as ImagePullSecret or
+in any other proprietary way).
+
+### Deep Dive
+
+To scan a container image of a given K8s workload Starboard will create a corresponding container of a scan Job and
+override its entrypoint to invoke Trivy filesystem scanner.
+
+This approach requires Trivy executable to be downloaded and made available to the entrypoint. We'll do that by adding
```
kfox1111, line 39: This could maybe be done with a csi driver as well.

docs/design/design_trivy_file_system_scanner.md
```diff
@@ -0,0 +1,197 @@
...
+the init container to the scan Job. Such init container will use the Trivy container image to copy Trivy executable out
+to the emptyDir volume, which will be shared with the other containers.
+
+Another init container is required to download Trivy vulnerability database and save it to the mounted shared volume.
```
kfox1111, line 43: Same.

docs/design/design_trivy_file_system_scanner.md
```diff
@@ -0,0 +1,197 @@
...
+filesystem scan. (See the provided [Example](#example) to have a better idea of all containers defined by a scan Job and
+how they share data via the emptyDir volume.)
+
+> Note that the second init container is required in [Standalone] mode, which is the only mode supported by Trivy
+> filesystem scanner at the time of writing this proposal.
+
+We further restrict scan Jobs to run on the same node where scanned Pod is running and never pull images from remote
+registries by setting the `ImagePullPolicy` to `Never`. To determine the node for a scan Job Starboard will list active
+Pods controlled by the scanned workload. If the list is not empty it will take the node name from the first Pod,
+otherwise it will ignore the workload.
```
kfox1111, line 55: This may run afoul of this admission controller https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#alwayspullimages

danielpacak: That's a very good point. We've merge this PR already but I opened #900 to add you comments as remark.

docs/design/design_trivy_file_system_scanner.md
danielpacak: I'm not sure that I understood this comment. Could you elaborate on how we can use csi driver in this case?

kfox1111: just thinking, instead of copying the file from an init container to an emtpydir, a csi driver in ephemeral mode could mount the file into the container.

kfox1111: May not be worth the effort. but could be like:

kind: Pod
apiVersion: v1
metadata:
  name: my-csi-app-inline-volume
spec:
  containers:
    - name: my-frontend
      image: busybox
      command: [ "sleep", "100000" ]
      volumeMounts:
      - mountPath: "/trivy"
        name: my-csi-volume
  volumes:
  - name: my-csi-volume
    csi:
      driver: trivy

PR #844, opened 2021-12-08
(fix) Upgrade starboard base container CVE-2021-3711,CVE-2021-36159.

goroi:
Upgrade starboard base container CVE-2021-3711,CVE-2021-36159.


danielpacak (approved):
Thank you @goroi This is looking good!

PR #845, opened 2021-12-09
ci(lint) : setup golangci-lint

mmorel-35:


.github/workflows/publish-helm-chart.yaml
```diff
@@ -16,22 +16,22 @@ jobs:
     runs-on: ubuntu-20.04
     steps:
       - name: Checkout
-        uses: actions/checkout@ec3a7ce113134d7a93b817d10a8272cb61118579 #v2.4.0
+        uses: actions/checkout@ec3a7ce113134d7a93b817d10a8272cb61118579  # v2.4.0
```
danielpacak, line 19: Why adding golangci-linter has anything to do with Helm chart workflow?

.github/workflows/golangci-lint.yml
```diff
@@ -0,0 +1,23 @@
+name: Golangci-lint
```
danielpacak, line 1: We already have a workflow to do static analysis on of source code on each PR defined in `.github/workflow/build.yaml`
Instead of adding another YAML, could we run golangci-lint action as a step under the `verify-code` job? This way we'll group checks and workflow triggers in one place.

danielpacak (changes requested):
Thank you for working on this PR @mmorel-35 Overall it looks good to me. I only suggested reusing existing workflow definition file to add linter step.

Please squash and sign-off your commits as well.

.github/workflows/publish-helm-chart.yaml
mmorel-35: It is a fix for ibiqlik/action-yamllint@v3 warnings

danielpacak: Ah I see. Please check also my comment about golangci-lint.yml that you added and reusing build.yaml instead

.github/workflows/build.yaml
```diff
@@ -218,12 +218,11 @@ jobs:
           go-version: ${{ env.GO_VERSION }}
       - name: Checkout code
         uses: actions/checkout@v2
-      - uses: actions/cache@v2
+      - name: Setup golangci-lint
+        uses: golangci/golangci-lint-action@v2
```
mmorel-35, line 222: There is already a cache with golangci-lint-action

danielpacak: Is it shared between steps? We run go mod vendor before we do verify code generated by clientgo-gen

danielpacak: Also if we reshuffle all these steps one day I'd prefer keeping the cached action config explicit.

danielpacak (approved):
Thanks for addressing all the comments @mmorel-35 🚀

Issue #537, opened 2021-04-26
feat: vulnerability rescan of K8s workloads based on report TTL


danielpacak:
This is a placeholder to specify how such functionality can be implemented

glimpsec:
A nightly rescan, specifically with Trivy, would help us recognise newly discovered CVE's in production workloads. If Trivy only scans on pod change, older pods won't have a recent scan. Old pods have an increased risk of running vulnerable images.

We're excited about the Starboard project, and this feature would make us reconsider it for production use.

kfox1111:
This would be useful. when we started playing around, we noticed that deleting the VulnerabilityReport object caused it to immediately rescan when using the operator. Maybe all that is needed is a cronjob that deletes all VulnerabilitieReports.

henrysachs:
I know this is kind of bad practice but is there any update on the plans of this? The issue is open since april and I just wanted to know if this is still on the "roadmap". If so what would be the best way to implement this? Maybe i can kind of support right here?

danielpacak:
Just to update readers on a current status we are working on a few enhancements to vulnerability scanning to make it more efficient. This may change the way we persist VulnerabilityReport instances and impact a lifecycle of each report. On the other hand, such braking changes will be probably introduced behind feature gates. What's more, the mechanism for rescanning could be agnostic to the lifecycle of VulnerabilityReport object.

Therefore, if you want to help with general rescanning concept we'd appreciate any high level overview for the solution that's compatible with K8s controller pattern and current architecture of Starboard. We'll take it from there.

kfox1111:
+1

Is there a process that we can use to help? Something similar to a KEP? Then we could comment on the general design and help enhance it.

danielpacak:
We haven't introduced any formal proposal workflow at this stage such as KEP. We can transfer this issue to discussion and start from there. Public GDoc with a few diagrams would work as well.

I know we weren't consistent but I'd prefer discussion design and idea brainstorming under GitHub discussion. Issues are good for bugs and finite well defined tasks. Otherwise we'll keep them opened for way too long.

kfox1111:
GDoc would allow for inline comments, which may be helpful for quicker iteration. Maybe we work there for a little while and then bring it to discussions?

danielpacak:
GDoc for getting started would do the job. However, for other design tasks we decided (last week) to host them under /docs/design

NissesSenap:
[...]

PR #863, opened 2021-12-19
docs: Design proposal for TTL scans

NissesSenap:
A start for a design proposal for ttl on scans #537 

Signed-off-by: Edvin Norling <edvin.norling@xenit.se>

### DIFF
--- a/docs/design/ttl_scans.md
+++ b/docs/design/ttl_scans.md
@@ -1,0 +1,74 @@
+# TTL scans
+
+## Summary
+
+Add an option to automatically delete old security reports. In this first version focus on vulnerability reports but in the long run we could add similar functionality to other reports as well.
+
+## Motivation
+
+In [537](https://github.com/aquasecurity/starboard/issues/537) we talk about a need to run nightly vulnerability scans of CVE:s.
+This way we can make sure to get new CVE reports for long time running pods as well.
+
+## Proposal
+
+Add a environment variable to the operator, for example `OPERATOR_VULNERABILITY_SCANNER_REPORT_TTL=86400`, this way we can add other ttl values for other reports as well.
+Adding this environment variable would add a annotation to the generated VulnerabilityReport that we can look for.
+
+Create a new controller that looks for changes in vulnerabilityreports and uses [RequeueAfter](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/reconcile#Result) calculated from the TTL annotation.
+At startup the operator will look through all existing vulnerabilityreports and delete existing ones where TTL have expired.
+If the TTL haven't expired the the vulnerabilityreports will be requeued and automatically checked again when the TTL have expired.
+
+We could calculate the ttl without having creating a new annotation to the reports but the verbosity of showing the users how long each report
+got a ttl outweighs the "issue" of generating a new annotation.
+
+### Example
+
+Below you can see a shortened version of the yaml. Notice the `metadata.annotations.starboard.aquasecurity.github.io/report-ttl` which is new.
+The operator would automatically apply the `starboard.aquasecurity.github.io/report-ttl` annotation to all new reports that it generates assuming that the environment variable is set.
+In theory users could also extend the TTL manually for a specific report by changing the starboard.aquasecurity.github.io/report-ttl annotation per VulnerabilityReport.
+
+```vulnerabilityReport.yaml
+apiVersion: aquasecurity.github.io/v1alpha1
+kind: VulnerabilityReport
+metadata:
+  creationTimestamp: "2021-12-08T12:03:48Z"
+  annotations:
+    starboard.aquasecurity.github.io/report-ttl: 24h
+  labels:
+    resource-spec-hash: 86b58dcb99
+    starboard.container.name: manager
+    starboard.resource.kind: ReplicaSet
+    starboard.resource.name: source-controller-b5d5cfdf4
+    starboard.resource.namespace: flux-system
+  name: replicaset-source-controller-b5d5cfdf4-manager
+report:
+  artifact:
+    repository: fluxcd/source-controller
+    tag: v0.16.1
+  registry:
+    server: ghcr.io
+  scanner:
+    name: Trivy
+    vendor: Aqua Security
+    version: 0.19.2
+  summary:
+    criticalCount: 0
+    highCount: 0
+    lowCount: 0
+    mediumCount: 0
+    unknownCount: 0
+  updateTimestamp: "2021-12-08T12:03:48Z"
+  vulnerabilities: []
+```
+
+Another option is to define a new report entry, the positive thing with that is that we can define the input type, in our case a `time.Duration`.
+
+## Alternatives
+
+Another "simpler" option could be to add the same environment variable `OPERATOR_VULNERABILITY_SCANNER_REPORT_TTL=2h` but instead of using RequeueAfter in the controller we could create a cronjob/job that runs once an hour and check for the same annotation.
+
+The bad thing hear is that we would have to manage yet another cronjob/job. We would also have to mange a new binary feature flag to run in cronjob cleanup mode.
+It would also trigger removal of multiple reports at the same time, compared to the event driven solution that would be much more precise per report
+and thus spreading out the new reports more.
+
+But the good thing is that everyone knows how jobs/cronjobs works especially since it's already well used within the starboard operator.


docs/design/ttl_scans.md
```diff
@@ -0,0 +1,31 @@
...
+
+## Motivation
+
+In [537](https://github.com/aquasecurity/starboard/issues/537) we talk about a need to run nightly vulnerability scans of CVE:s.
+This way we can make sure to get new CVE reports for long time running pods as well.
+
+## Proposal
+
+Add a feature feature flag of the operator, for example `--ttlVulnerabilityReport 24h`, this way we can add other ttl values for other reports as well.
+Adding this flag would add a spec/annotation to the generated VulnerabilityReport that we can look for.
```
danielpacak, line 15: Each VulnerabilityReport has `metadata.creationTimestamp` and `report.updateTimestamp`. Do you think we could use these properties to determine TTL without adding annotations? On the other hand the annotation is very verbose / explicit to indicate that this report has validity timestamp.

docs/design/ttl_scans.md
```diff
@@ -0,0 +1,31 @@
...
+
+Using similar functionality as defined in [ttlAfterFinished](https://kubernetes.io/docs/concepts/workloads/controllers/ttlafterfinished/) for cronjob seems to be the most versatile option.
+
+Just like the [ttlAfterFinished controller](https://github.com/kubernetes/kubernetes/blob/2a821d787b7f21fbb399b0a0bf6d3fa0c79f4293/pkg/controller/ttlafterfinished/ttlafterfinished_controller.go) we should add it as a separate controller.
```
danielpacak, line 19: Just to confirm that we are talking about VulnerabilityReport controller here, right? And later on we could watch other kinds of security reports in a similar way.

docs/design/ttl_scans.md
```diff
@@ -0,0 +1,31 @@
...
+This way it will be easy to extend to other reports as well.
+
+## Alternatives
+
+Another "simpler" option could be to add the same feature flag `--ttlVulnerabilityReport 24h` but instead of doing a similar solution as defined in ttlAfterFinished we could create a cronjob/job that runs once an hour and check for the same annotation.
```
danielpacak, line 24: I really like the "event driven" approach described in the proposal. Especially to avoid deleting multiple VulnerabilityReports around the same time, which would create multiple (re)scan Jobs. Not sure how feasible is that, but more jitter in delete events the better overall.

With a CronJob we'd need to create it programmatically or add it as part of the installation descriptors. We'd also need a container and command to run it.

danielpacak (commented):
Thank you for working on this proposal @NissesSenap . For me it looks solid and we could implement and try out easily. I just left a few comments for clarity.

docs/design/ttl_scans.md
kfox1111: I think thats whats happening with kubernetes job type listed below. Following their pattern would be good. Adding a ttl 24h annotation, label, or spec entry, then comparing the current type with report.updateTimestamp + the ttl?

docs/design/ttl_scans.md
kfox1111: That may or may not be seen as a good thing depending on the user.

Some users might want to know that starting at some point in time (daily, weekly, whatever) that all scans have happened recently.

The operator one way or another will need to be stable in the face of a lot of new scan requests, so potentially triggering a lot of rescans needs to be handled anyway I think.

I guess the event based one (24h + age of object) can still be pretty easily be turned into the latter by just hard deleting all the reports with a cron job and skipping the TTL feature by an end user. So they are not mutually exclusive.

docs/design/ttl_scans.md
kfox1111: Yeah. I think the idea here is to have one ttl controller for all reports, not taking each existing controller and adding ttl to them.

NissesSenap: Just like @kfox1111 writes this is a seperate controller. For now I think we should focus on VulnerabilityReport but In the long run I think this might be useful for other reports as well. For example: https://github.com/aquasecurity/starboard/pull/830

This is also why my idea is to use `--ttlVulnerabilityReport 24h` and not `ttlReport 24h` since then we will be able to configure different values for different reports.

NissesSenap:
I have tried to make the explanation more verbose following your comments @danielpacak and @kfox1111.
Stopped talking about feature flags and instead talk about environment variable since this is something that is used today.
Also given a short example in the yaml to make it super clear.

docs/design/ttl_scans.md
```diff
@@ -0,0 +1,79 @@
...
+If the operator restarts the queue will disappear since we don't hold any state so we would have to scan all the available reports and calculate
+the ttl comparing it to `report.updateTimestamp`.
+
+We could calculate the ttl without having creating a new annotation to the reports but the verbosity of showing the users how long each report
+got a ttl outweighs the "issue" of generating a new annotation.
+
+### Example
+
+Below you can see a shortened version of the yaml. Notice the `metadata.annotations.ttlVulnerabilityReport` which is the new.
+The operator would automatically apply the `ttlVulnerabilityReport` annotation to all reports that it generates assuming that the environment variable is set.
```
danielpacak, line 31: According to [Syntax and character set](https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/#syntax-and-character-set) an annotation has prefix and name to avoid collisions.

Maybe we could use `starboard.aquasecurity.github.io/report-ttl` for any kind of security report managed by Starboard. WDYT?

docs/design/ttl_scans.md
```diff
@@ -0,0 +1,79 @@
...
+
+Add a environment variable to the operator, for example `OPERATOR_TTL_VULNERABILITY_SCANNER=24h`, this way we can add other ttl values for other reports as well.
+Adding this environment variable would add a annotation to the generated VulnerabilityReport that we can look for.
+
+Using similar functionality as defined in [ttlAfterFinished](https://kubernetes.io/docs/concepts/workloads/controllers/ttlafterfinished/) for cronjob seems to be the most versatile option.
+
+Just like the [ttlAfterFinished controller](https://github.com/kubernetes/kubernetes/blob/2a821d787b7f21fbb399b0a0bf6d3fa0c79f4293/pkg/controller/ttlafterfinished/ttlafterfinished_controller.go) we should add it as a separate controller.
+This way it will be easy to extend to other reports as well.
+
+If the operator restarts the queue will disappear since we don't hold any state so we would have to scan all the available reports and calculate
```
danielpacak, line 22: We are using controller-runtime and its Reconciler abstraction. Can you elaborate how we're going to init the controller on startup? If we block the init for too longit won't scale with a number of VulnerabilityReports. On the other hand, I'm not sure we have "access" to work queue outside the Reconilation handler function.

danielpacak (commented):
Great progress @NissesSenap I left a few more comments before we approve your proposal. Could you sing the SLA, which is required to merge the PR. Thank you!

docs/design/ttl_scans.md
NissesSenap: I know about this but for some reason I didn't think about it when writing the example... good catch and I have updated the PR accordingly.

docs/design/ttl_scans.md
```diff
@@ -0,0 +1,79 @@
...
+Add an option to automatically delete old starboard scans, in this first version focus on vulnerability scans but in the long run we could add similar functionality to other reports as well.
+
+## Motivation
+
+In [537](https://github.com/aquasecurity/starboard/issues/537) we talk about a need to run nightly vulnerability scans of CVE:s.
+This way we can make sure to get new CVE reports for long time running pods as well.
+
+## Proposal
+
+Add a environment variable to the operator, for example `OPERATOR_TTL_VULNERABILITY_SCANNER=24h`, this way we can add other ttl values for other reports as well.
```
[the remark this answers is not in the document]
NissesSenap: Sounds reasonable, I went with OPERATOR_VULNERABILITY_SCANNER_REPORT_TTL to make it super clear that is about the reports.

docs/design/ttl_scans.md
NissesSenap: Spoken to a friend about this and he pointed out that we don't need to use the ttlAfterFinished solution.
Instead we can use the controller result. https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/reconcile#Result

When we get a reconcile thanks to a vulnerability have been created we can go in and check the annotation `starboard.aquasecurity.github.io/report-ttl` and do the calculation on how long it should live.
and then just do a RequeueAfter and the controller will manage this magic for us :) 

```.go
type Result struct {
	// Requeue tells the Controller to requeue the reconcile key.  Defaults to false.
	Requeue bool

	// RequeueAfter if greater than 0, tells the Controller to requeue the reconcile key after the Duration.
	// Implies that Requeue is true, there is no need to set Requeue to true at the same time as RequeueAfter.
	RequeueAfter time.Duration
}
```

My understanding is that this also will be checked during startup of the controller if there is already existing reports.
Does this sound reasonable @danielpacak? If so I will update the document sometime after Christmas and see if I can create a PoC around it.

danielpacak: If we could use RequestAfter duration and confirm that existing VulnerabilityReports get listed by the controller manager that would be the best clean solution.

NissesSenap:
I have created a PoC https://github.com/aquasecurity/starboard/pull/879 using the RequeueAfter functinallity.
This seems to work fine at least on a small cluster. Even if your cluster is big I don't think this should create any more issues than starting to listen for all replicasets in a cluster.

As discussed earlier I use the annotation `starboard.aquasecurity.github.io/report-ttl` but I don't support adding a time granularity instead I always assume that the ttl defined is in seconds. We could of course add a feature where we listen for hours, minutes etc.

```.yaml
annotations:
  starboard.aquasecurity.github.io/report-ttl: 24h
```
instead of:

```.yaml
annotations:
  starboard.aquasecurity.github.io/report-ttl: 86400
```

Another option could also be to keep only supporting seconds and instead of using a annotation we could add a new spec entry to the VulnerabilityReport spec and set the type to int.

Please have a look at the update design when you have time @danielpacak @kfox1111 and look at the PoC implementation.

docs/design/ttl_scans.md
```diff
@@ -0,0 +1,78 @@
+# TTL scans
+
+## Summary
+
+Add an option to automatically delete old starboard scans, in this first version focus on vulnerability scans but in the long run we could add similar functionality to other reports as well.
```
danielpacak, line 5:
```suggestion
Add an option to automatically delete old security reports. In this first version focus on vulnerability reports but in the long run we could add similar functionality to other reports as well.
```

docs/design/ttl_scans.md
```diff
@@ -0,0 +1,78 @@
...
+In [537](https://github.com/aquasecurity/starboard/issues/537) we talk about a need to run nightly vulnerability scans of CVE:s.
+This way we can make sure to get new CVE reports for long time running pods as well.
+
+## Proposal
+
+Add a environment variable to the operator, for example `OPERATOR_VULNERABILITY_SCANNER_REPORT_TTL=86400`, this way we can add other ttl values for other reports as well.
+Adding this environment variable would add a annotation to the generated VulnerabilityReport that we can look for.
+
+Create a new controller that looks for changes in vulnerabilityreports and uses [RequeueAfter](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/reconcile#Result)
+to look through all existing vulnerabilityreports at startup of the operator but also monitor newly created reports and requeue them after need
```
danielpacak, line 18: I think that with RequeAfter property of reconciliation result we could avoid listing vulnerability reports at startup at all. The goal is to avoid batch processing as much as we can (especially on startup).

Therefore, I'd clarify that the annotation and the update timestamp is used to determine a TTL for a single report. If the report should be kept we adding it back to the work queue. Otherwise it's deleted.

docs/design/ttl_scans.md
```diff
@@ -0,0 +1,78 @@
...
+
+Using similar functionality as defined in [ttlAfterFinished](https://kubernetes.io/docs/concepts/workloads/controllers/ttlafterfinished/) for cronjob seems to be the most versatile option.
```
danielpacak, line 20:
```suggestion
Using similar functionality as defined in [ttlAfterFinished](https://kubernetes.io/docs/concepts/workloads/controllers/ttlafterfinished/) for Jobs seems to be the most versatile option.
```

danielpacak (changes requested):
@NissesSenap Lets clarify one last thing in this design proposal. Whether we do some (batch) processing on startup or not. Based on my tests and experiments we can implemented everything with RequeAfter property, but I'd like others to confirm.

NissesSenap:
Is it a good idea to use annotation or should we think about using the status field instead?
This could help us with doing less type conversion (not that it's a huge problem) but it solves some other potential issues as well, for example users won't be able to update the TTL value manually.
https://github.com/aquasecurity/starboard/pull/879#discussion_r782096294

danielpacak:
> Is it a good idea to use annotation or should we think about using the status field instead?
> This could help us with doing less type conversion (not that it's a huge problem) but it solves some other potential issues as well, for example users won't be able to update the TTL value manually.
> [#879 (comment)](https://github.com/aquasecurity/starboard/pull/879#discussion_r782096294)

I believe that the annotation is good to experiment and prove that this solution makes sense and is being adopted. This is what Kubernetes does as well. Once we feel comfortable we can refactor from annotation to typed property.

docs/design/ttl_scans.md
NissesSenap: Rewrote the section and I hope It made it clearer.

NissesSenap:
I have rewritten the design document to match more the final implementation that we went for.
Remove old mentions of the old idea.

docs/design/ttl_scans.md
danielpacak: Let's see how the event drive approach is adapter. We can always think of enhancement and add CronJob if it's really needed.

danielpacak (approved):

PR #872, opened 2021-12-23
feat: Vulnerability scan using Trivy fs scan command

deven0t:
``trivy.command: fs`` will change the trivy scan option to pick up vulnerability scans
using trivy fs scan command

This is the implementation of first approach suggested here #830

### DIFF
--- a/pkg/kube/object.go
+++ b/pkg/kube/object.go
@@ -300,6 +300,8 @@ func (o *ObjectResolver) ObjectFromObjectRef(ctx context.Context, workload Objec
 }
 
 var ErrReplicaSetNotFound = errors.New("replicaset not found")
+var ErrNoRunningPods = errors.New("no active pods for controller")
+var ErrUnSupportedKind = errors.New("unsupported workload kind")
 
 // ReportOwner resolves the owner of a security report for the specified object.
 func (o *ObjectResolver) ReportOwner(ctx context.Context, obj client.Object) (client.Object, error) {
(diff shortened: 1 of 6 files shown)


pkg/kube/object.go
```diff
@@ -445,6 +445,59 @@ func (o *ObjectResolver) GetRelatedReplicasetName(ctx context.Context, object Ob
...
+	switch t := obj.(type) {
+	case *corev1.Pod:
+		return (obj.(*corev1.Pod)).Spec.NodeName, nil
+	case *appsv1.Deployment:
+		replicaSet, err := o.ReplicaSetByDeployment(ctx, obj.(*appsv1.Deployment))
+		if err != nil {
+			return "", err
+		}
+		pods, err := o.getPodsByLabelSelector(ctx, obj.GetNamespace(), replicaSet.Spec.Selector.MatchLabels)
+		if err != nil || (pods != nil && len(pods.Items) == 0) {
```
danielpacak, line 459: I think we should distinguish error returned by client-go (due to connection problems with K8s API server) and error returned because replicas count is 0.

And actually, I'm not sure that we should throw an error in case there are no pods.

pkg/kube/object.go
```diff
@@ -488,3 +541,15 @@ func (o *ObjectResolver) getReplicaSetByPod(ctx context.Context, object Object)
 	}
 	return controller.Name, nil
 }
+
+func (o *ObjectResolver) getPodsByLabelSelector(ctx context.Context, namespace string,
```
danielpacak, line 545: Returned PodList is just a wrapper for []Pod so I'd just return []Pod.

pkg/vulnerabilityreport/builder.go
```diff
@@ -68,11 +70,21 @@ func (s *ScanJobBuilder) WithCredentials(credentials map[string]docker.Auth) *Sc
...
 func (s *ScanJobBuilder) Get() (*batchv1.Job, []*corev1.Secret, error) {
 	spec, err := kube.GetPodSpec(s.object)
 	if err != nil {
 		return nil, nil, err
 	}
+	podSpecHash := kube.ComputeHash(spec)
+
+	// get nodename from running pods.
+	resolver := kube.ObjectResolver{Client: s.client}
+	spec.NodeName, err = resolver.GetNodeName(context.Background(), s.object)
```
danielpacak, line 87: Shouldn't we do that only if trivy fs is enabled? I'm not sure that the job builder is the right place to resolve node name anyway. I'd do that within Trivy plugin, which returns scan job spec with node name (if required by config).

pkg/operator/controller/vulnerabilityreport.go
```diff
@@ -228,8 +230,17 @@ func (r *VulnerabilityReportReconciler) submitScanJob(ctx context.Context, owner
 		WithTolerations(scanJobTolerations).
 		WithAnnotations(scanJobAnnotations).
 		WithCredentials(credentials).
+		WithClient(r.Client).
 		Get()
 
+	if err != nil {
+		if err.Error() == "cannot get nodename from podspec" {
```
danielpacak, line 237: We should either use global variable to do such comparisons (e.g. kube.ErrUnknownNodeName) or return blank string when we resolve node name. Anyway, we should probably move this logic away and retrieve node name in Trivy plugin impl.

danielpacak (changes requested):
Hey @deven0t It's a good start. See my comments and let me know wdyt?

pkg/kube/object.go
deven0t: I checked on this, in getPodsbyLabelSelector method we use list method from client, so when there are no records matching those labels then we dont get any error and pods.Items is empty slice. so error is nil and len(pods.Items) will be zero. So we are not throwing any error when there are no pods. 
I think thats why I added a check with pods != nil && len(pods.Items) == 0, so that we will not get nil pointer error.

btw this will be needed for err check after ReplicaSetByDeployment(), so we will not throw error when there are no replicasets.

pkg/operator/controller/vulnerabilityreport.go
```diff
@@ -4,6 +4,7 @@ import (
 	. "github.com/aquasecurity/starboard/pkg/operator/predicate"
 
 	"context"
+	errors2 "errors"
```
danielpacak, line 7: I think it's better to use alias for "k8s.io/apimachinery/pkg/api/errors", e.g.:

```go
import (
	"errors"

	apierrors "k8s.io/apimachinery/pkg/api/errors"
	// or
	k8sapierrors "k8s.io/apimachinery/pkg/api/errors"
)
```

pkg/operator/controller/vulnerabilityreport.go
```diff
@@ -242,6 +245,14 @@ func (r *VulnerabilityReportReconciler) submitScanJob(ctx context.Context, owner
 		WithCredentials(credentials).
 		Get()
 
+	if err != nil {
+		if errors2.Is(err, kube.ErrNodeNameNotFound) {
+			log.V(1).Info("Cannot run a scan job as nodeName is not available")
```
danielpacak, line 250: We should explicitly state in logs that we're ignoring a workload scan because there are no pods (= 0 replicas). "nodeName is not available" is pretty scary and may mislead to infrastructure problems rather than deliberate logic implemented by Starboard.

pkg/plugin/trivy/plugin.go
```diff
@@ -862,6 +909,212 @@ func (p *plugin) getPodSpecForClientServerMode(ctx starboard.PluginContext, conf
...
+//
+func (p *plugin) getPodSpecForStandaloneFSMode(ctx starboard.PluginContext, config Config,
+	workload client.Object) (corev1.PodSpec, []*corev1.Secret, error) {
+	var secrets []*corev1.Secret
+	spec, err := kube.GetPodSpec(workload)
+	if err != nil {
+		return corev1.PodSpec{}, nil, err
+	}
+	// get nodeName from running pods.
+	spec.NodeName, err = p.objectResolver.GetNodeName(context.Background(), workload)
```
danielpacak, line 926: I think we should reuse context.Context passed from Starboard to the plugin instead of instantiating a new one.

pkg/plugin/trivy/plugin.go
```diff
@@ -862,6 +909,212 @@ func (p *plugin) getPodSpecForClientServerMode(ctx starboard.PluginContext, conf
...
+
+	if spec.NodeName == "" {
+		return corev1.PodSpec{}, nil, kube.ErrNodeNameNotFound
```
danielpacak, line 929: TBD if this error should be returned by the GetNodeName method.

danielpacak (commented):
👋 @deven0t Overall the changes look good to me! I left a few more comments to consider. Let's catch up and discuss before we merge this PR.

danielpacak (approved):

PR #878, opened 2021-12-27
enable more linters and flag codecov reports

mmorel-35:


itest/helper/helper.go
```diff
@@ -245,13 +245,11 @@ func (b *DeploymentBuilder) Build() *appsv1.Deployment {
...
 }
 
-var (
-	trivyScanner = v1alpha1.Scanner{
-		Name:    "Trivy",
-		Vendor:  "Aqua Security",
-		Version: "0.16.0",
-	}
-)
+var trivyScanner = v1alpha1.Scanner{
```
danielpacak, line 248: No sure we want to be so strict about linting Go code. At some point we are probably going to spend most of the time making linter happy, whereas such things are not very relevant and super critical. I'd just enable linters that are supported by popular IDEs and check for real problems.

pkg/kube/object.go
```diff
@@ -303,17 +303,17 @@ var ErrReplicaSetNotFound = errors.New("replicaset not found")
 
 // ReportOwner resolves the owner of a security report for the specified object.
 func (o *ObjectResolver) ReportOwner(ctx context.Context, obj client.Object) (client.Object, error) {
-	switch obj.(type) {
+	switch obj := obj.(type) {
 	case *appsv1.Deployment:
-		return o.ReplicaSetByDeployment(ctx, obj.(*appsv1.Deployment))
+		return o.ReplicaSetByDeployment(ctx, obj)
```
danielpacak, line 308: Does this even work without explicit cast?

danielpacak (changes requested):
Thank you for working on this patch @mmorel-35 , but I'm quite sceptical when it comes to enabling so many linters. They are good for catching basic issues, but once we enable too many of them we may suffer merging simplest PRs because of some issues that are not captured by go vet or popular IDEs. I'd rather run all such linters on demand (or periodically) and fix most critical problems they may find by opening a dedicated PR. WDYT?

danielpacak:
Closing as these are not real issues that must be addressed immediately.

PR #879, opened 2021-12-29
TTL scan vulnerability report 

NissesSenap:
This controller is used to manage TTL of vulnerabilityreports.
If the ttl annotation passes it automatically deletes the report.

In the long run this controller could be used for other reports as well, but for now we will focus on vulnerabilityreports.

This is a PoC to continue the discussion around https://github.com/aquasecurity/starboard/pull/863.
If we think this is a okay way forward I hope we can use this PR to work towards solving: https://github.com/aquasecurity/starboard/issues/537

There are a number of TODO:s in the code and the k8s yaml needs to be added. There is probably other things as well but hopefully it's a good base to stand on.

Currently the controller only looks for a int and uses them as seconds when setting the TTL. We need to decide if we should support other values then seconds.


NissesSenap:
The code isn't perfect, I'm aware of the issue in the test but I didn't want to keep on working on it unless we think this is the way forward.

pkg/operator/controller/ttl_report.go
```diff
@@ -0,0 +1,87 @@
...
+)
+
+type TTLReportReconciler struct {
+	logr.Logger
+	etc.Config
+	client.Client
+}
+
+func (r *TTLReportReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	err := ctrl.NewControllerManagedBy(mgr).
```
danielpacak, line 25: NIT: We can just return the result of ctrl.NewControllerManagerBy().Complete() here if we don't wrap the error

pkg/operator/controller/ttl_report.go
```diff
@@ -0,0 +1,87 @@
...
+		report := &v1alpha1.VulnerabilityReport{}
+		err := r.Client.Get(ctx, req.NamespacedName, report)
+		if err != nil {
+			if errors.IsNotFound(err) {
+				log.V(1).Info("Ignoring cached report that must have been deleted")
+				return ctrl.Result{}, nil
+			}
+			return ctrl.Result{}, fmt.Errorf("getting report from cache: %w", err)
+		}
+		ttlReportAnnotationStr := report.Annotations[v1alpha1.TTLReportAnnotation]
```
danielpacak, line 47: I think we may want to distinguish cases when the annotation is not set (TTL controlled disabled) from scenario when it is set but has no value (data consistency error that's only possible if some edited the vulnerability report instances manually).

pkg/operator/controller/ttl_report.go
```diff
@@ -0,0 +1,87 @@
...
+		if ttlReportAnnotationStr != "" {
+			creationTime := report.GetCreationTimestamp()
```
danielpacak, line 49: To persist vulnerability reports we do create or update so in general we should be looking at the UpdateTimestamp property. WDYT?

```suggestion
			creationTime := report.Report.UpdateTimestamp
```
BTW, in createOrUpdate method we do clone instance without rewriting annotations.

pkg/operator/controller/ttl_report.go
```diff
@@ -0,0 +1,87 @@
...
+			ttlExpired, durationToTTLExpiration, err := ttlIsExpired(ttlReportAnnotationStr, creationTime.Time)
+			if err != nil {
+				// TODO add a specific test to check if the error is that we can't convert to string
+				// Should be a custom error that we can use and point the user whats wrong. This should never happen... We controle the annotation
+				return ctrl.Result{}, err
+			}
+			if ttlExpired {
+				log.V(1).Info("Removing vulnerabilityReport")
+				r.Client.Delete(ctx, report, &client.DeleteOptions{})
```
danielpacak, line 58: We should handle error here. In particular we should ignore NotFound errors, retry otherwise.

pkg/operator/controller/ttl_report.go
```diff
@@ -0,0 +1,87 @@
...
+				// TODO Should we add a small delay I get "Ignoring cached report that must have been deleted", directly after a deletion
+				return ctrl.Result{Requeue: true}, nil
```
danielpacak, line 60: IMO Once we delete the report we don't need to requiem. When new report is created we will have another chance to check its TTL. I'm wondering if you still get such error without requeue.

pkg/operator/controller/ttl_report.go
```diff
@@ -0,0 +1,87 @@
...
+			}
+			log.V(1).Info("RequeueAfter", "durationToTTLExpiration", durationToTTLExpiration)
+			return ctrl.Result{RequeueAfter: durationToTTLExpiration}, nil
+		}
+		return ctrl.Result{}, err
+	}
+}
+
+func ttlIsExpired(reportTTL string, creationTime time.Time) (bool, time.Duration, error) {
+	reportTTLInt, err := strconv.ParseInt(reportTTL, 10, 64)
```
danielpacak, line 70: By using `time.ParseDuration` we can easily support all duration literal (20s, 15m, 1m20s), not just seconds.

pkg/operator/etc/config.go
```diff
@@ -23,6 +23,7 @@ type Config struct {
 	HealthProbeBindAddress        string        `env:"OPERATOR_HEALTH_PROBE_BIND_ADDRESS" envDefault:":9090"`
 	CISKubernetesBenchmarkEnabled bool          `env:"OPERATOR_CIS_KUBERNETES_BENCHMARK_ENABLED" envDefault:"true"`
 	VulnerabilityScannerEnabled   bool          `env:"OPERATOR_VULNERABILITY_SCANNER_ENABLED" envDefault:"true"`
+	VulnerabilityScannerReportTTL string        `env:"OPERATOR_VULNERABILITY_SCANNER_REPORT_TTL" envDefault:""`
```
danielpacak, line 26: We can use time.Duration and it will get parsed automatically. See ScanJobTimeout property.

pkg/operator/controller/vulnerabilityreport.go
```diff
@@ -353,6 +353,8 @@ func (r *VulnerabilityReportReconciler) processCompleteScanJob(ctx context.Conte
 			Container(containerName).
 			Data(reportData).
 			PodSpecHash(podSpecHash).
+			//TODO this should only be done if TTL is not "", how to make nice?
+			VulnerabilityScannerReportTTL(r.Config.VulnerabilityScannerReportTTL).
```
danielpacak, line 357: One way could be:

```go
reportBuilder := vulnerabilityreport.NewReportBuilder(r.Client.Scheme()).
	Controller(owner).
	Container(containerName).
	Data(reportData).
	PodSpecHash(podSpecHash)

if r.Config.VulnerabilityScannerReportTTL != "" {
	reportBuilder.VulnerabilityScannerReportTTL(r.Config.VulnerabilityScannerReportTTL)
}

report, err := reportBuilder.Get()
```

danielpacak (changes requested):
Great job @NissesSenap ! I tested your branch and it seems to be working as expected. I also did a quick test for scenario where I scaled down the operator while some keys were in the queue, then I scaled it up, and it was able to catch up with events and delete a report according to its TTL.

Please check my comments and suggestions so we can proceed and merge your PR.

pkg/operator/operator.go
```diff
@@ -160,6 +160,14 @@ func Start(ctx context.Context, buildInfo starboard.BuildInfo, operatorConfig et
 		}).SetupWithManager(mgr); err != nil {
 			return fmt.Errorf("unable to setup vulnerabilityreport reconciler: %w", err)
 		}
+
+		if err = (&controller.TTLReportReconciler{
```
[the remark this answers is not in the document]
NissesSenap: It's currently inside the if `operatorConfig.VulnerabilityScannerEnabled` so that is already checked. But we could ask our self if that is a good thing or not. For now it works but if we want the controller to manage multiple reports in the future we will need to move it out.
But it's a bit of pre-optimization to do that and for now I think we can skip it.

NissesSenap:
Thanks for giving it a try @danielpacak and I'm glad you like it.
I will continue working on this PR and the coming days.

pkg/operator/controller/ttl_report.go
NissesSenap: Good point, I now check if the annotation is empty and then return an error.
And then continue down as earlier.

pkg/operator/controller/ttl_report.go
NissesSenap: Hadn't actually looked what UpdateTimestamp did. It sounds reasonable to change over to it.

pkg/operator/controller/ttl_report.go
NissesSenap: Sadly I still get the error, I think that the controller is just to quick and even if I now have added `predicate.IsBeingTerminated` and actually remove not requeue just as you pointed out I still get the same error.
One way could be to just not log the issue, I created a new TODO where the log is actually created to not forget this.

pkg/vulnerabilityreport/builder.go
```diff
@@ -193,6 +199,9 @@ func (b *ReportBuilder) Get() (v1alpha1.VulnerabilityReport, error) {
 
 	report := v1alpha1.VulnerabilityReport{
 		ObjectMeta: metav1.ObjectMeta{
+			Annotations: map[string]string{
```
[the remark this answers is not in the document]
NissesSenap: So true, just threw it in there during the PoC, added a if statment to handle it now.

pkg/operator/controller/ttl_report.go
NissesSenap: I started using `predicate.InstallModePredicate` to see if that would help with some things. Thanks to that I think it's good to save the error return.
What do you think?

NissesSenap:
One thing hits me when I'm playing around with this feature is that how should we help our users start using the feature?
If i start the operator today and I have a bunch of old reports in my cluster I get:

```
{"level":"error","ts":1641552666.1590643,"logger":"controller.vulnerabilityreport","msg":"Reconciler error","reconciler group":"aquasecurity.github.io","reconciler kind":"VulnerabilityReport","name":"replicaset-minio-5c8bcf5d48-minio","namespace":"default","error":"the annotation starboard.aquasecurity.github.io/report-ttl is missing from the report","stacktrace":"sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2\n\t/home/edvin/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.10.3/pkg/internal/controller/controller.go:227"}
```

We could of course solve this with code and delete all the reports that don't have the `annotation starboard.aquasecurity.github.io/report-ttl` but I can see why people don't think that is optimal.

A simple way for us could be to just write a migration guide where we write a small bash loop ether annotates all existing reports or we just delete them all and the operator will do the rest.

No matter which solution we pick we need to document this but I don't know where: `docs/operator/getting-started.md` don't feel like correct place for this.
`docs/operator/configuration.md` could be an option but with a new header.
Should we create a new document?

What do you think @danielpacak?

danielpacak:
> One thing hits me when I'm playing around with this feature is that how should we help our users start using the feature? If i start the operator today and I have a bunch of old reports in my cluster I get:
> 
> ```
> {"level":"error","ts":1641552666.1590643,"logger":"controller.vulnerabilityreport","msg":"Reconciler error","reconciler group":"aquasecurity.github.io","reconciler kind":"VulnerabilityReport","name":"replicaset-minio-5c8bcf5d48-minio","namespace":"default","error":"the annotation starboard.aquasecurity.github.io/report-ttl is missing from the report","stacktrace":"sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2\n\t/home/edvin/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.10.3/pkg/internal/controller/controller.go:227"}
> ```
> 
> We could of course solve this with code and delete all the reports that don't have the `annotation starboard.aquasecurity.github.io/report-ttl` but I can see why people don't think that is optimal.
> 
> A simple way for us could be to just write a migration guide where we write a small bash loop ether annotates all existing reports or we just delete them all and the operator will do the rest.
> 
> No matter which solution we pick we need to document this but I don't know where: `docs/operator/getting-started.md` don't feel like correct place for this. `docs/operator/configuration.md` could be an option but with a new header. Should we create a new document?
> 
> What do you think @danielpacak?

Starboard is incubating and w are not providing (yet) an official automated upgrade experience. Hence, we could just mention any manual steps on the GitHub release page or the [Upgrade](https://aquasecurity.github.io/starboard/v0.13.2/operator/installation/upgrade/) documentation page.

To handle vulnerability reports generated by the previous release of Starboard I'd leave them intact. And instruct user that they can delete them manually (which will trigger rescan) or patch by adding a TTL annotation, which will trigger the TTL controller.

pkg/operator/operator.go
```diff
@@ -160,6 +161,16 @@ func Start(ctx context.Context, buildInfo starboard.BuildInfo, operatorConfig et
 		}).SetupWithManager(mgr); err != nil {
 			return fmt.Errorf("unable to setup vulnerabilityreport reconciler: %w", err)
 		}
+
+		if operatorConfig.VulnerabilityScannerReportTTL != time.Duration(0) {
```
NissesSenap, line 165: Is it worth adding a function to perform the `operatorConfig.VulnerabilityScannerReportTTL != time.Duration(0) `? We perform it in two files and import time in multiple locations.

We could add this function in to: `pkg/operator/etc/config.go`. Then it would also be easy to write a small test for the function.

pkg/operator/controller/ttl_report.go
```diff
@@ -0,0 +1,95 @@
...
+
+func (r *TTLReportReconciler) reconcileReport() reconcile.Func {
+	return func(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+		log := r.Logger.WithValues("report", req.NamespacedName)
+
+		report := &v1alpha1.VulnerabilityReport{}
+		err := r.Client.Get(ctx, req.NamespacedName, report)
+		if err != nil {
+			if errors.IsNotFound(err) {
+				log.V(1).Info("Ignoring cached report that must have been deleted")
```
NissesSenap, line 50: Decided in the end that I think this Info is good to keep since it only shows during debugging of the app.

NissesSenap (commented):
Starting to feel ready for a review of this PR.
Have cleaned up the code from TODOS, added yaml config and created documentation.

Thanks for all your great comments @danielpacak

pkg/operator/controller/ttl_report.go
```diff
@@ -0,0 +1,95 @@
...
+		}
+		log.V(1).Info("RequeueAfter", "durationToTTLExpiration", durationToTTLExpiration)
+		return ctrl.Result{RequeueAfter: durationToTTLExpiration}, nil
+	}
+}
+
+func ttlIsExpired(reportTTL string, creationTime time.Time) (bool, time.Duration, error) {
+	reportTTLTime, err := time.ParseDuration(reportTTL)
+	if err != nil {
+		return false, time.Duration(0), err
```
danielpacak, line 83: Let's add more context to the error, e.g. `fmt.Errorf("failed parsing report TTL: %w")`. When we see it in the logs we'll know what went wrong without analysing stack traces.

pkg/operator/etc/config.go
```diff
@@ -27,6 +27,7 @@ type Config struct {
 	LeaderElectionEnabled                        bool          `env:"OPERATOR_LEADER_ELECTION_ENABLED" envDefault:"false"`
 	LeaderElectionID                             string        `env:"OPERATOR_LEADER_ELECTION_ID" envDefault:"starboard-lock"`
 	VulnerabilityScannerScanOnlyCurrentRevisions bool          `env:"OPERATOR_VULNERABILITY_SCANNER_SCAN_ONLY_CURRENT_REVISIONS" envDefault:"false"`
+	VulnerabilityScannerReportTTL                time.Duration `env:"OPERATOR_VULNERABILITY_SCANNER_REPORT_TTL" envDefault:"0s"`
```
danielpacak, line 30: If we used pointer to `time.Duration` we'd skip default values and we can check if it's on or off with `Config.VulnerabilityScannerReportTTL != nil`

```suggestion
	VulnerabilityScannerReportTTL                *time.Duration `env:"OPERATOR_VULNERABILITY_SCANNER_REPORT_TTL"`
```

pkg/vulnerabilityreport/builder.go
```diff
@@ -143,6 +143,7 @@ type ReportBuilder struct {
 	container  string
 	hash       string
 	data       v1alpha1.VulnerabilityReportData
+	reportTTL  string
```
danielpacak, line 146: Can we use *time.Duration instead of string here? To fail fast we should not pass invalid duration literals to the builder.

danielpacak (changes requested):
👋 @NissesSenap I left few more comments to address. Hope all make sense.

pkg/operator/controller/ttl_report.go
NissesSenap: I moved this outside of the function since I use time.Duration instead of string in ttlIsExpired

When doing time.ParseDuration i check for error and return:
```.go
		reportTTLTime, err := time.ParseDuration(ttlReportAnnotationStr)
		if err != nil {
		}
```

pkg/vulnerabilityreport/builder.go
NissesSenap: We could do that and manage the conversion from time.Duration to string inside the function. We need to do this thanks to that a annotation always is `map[string]string`.

If we don't want to do that we could also leave the idea of using a annotation and instead go over to using the status field to store the ttl data.

We could do it in a similar way that certmanager have done it in there status field.

```
apiVersion: cert-manager.io/v1
kind: Certificate
metadata:
  name: something
spec:
  commonName: something
  duration: 8760h0m0s
  issuerRef:
    name: intermediate-ca
  privateKey:
    algorithm: ECDSA
    size: 521
  revisionHistoryLimit: 1
  secretName: something
  usages:
  - server auth
  - client auth
status:
  conditions:
  - lastTransitionTime: "2021-06-21T10:42:17Z"
    message: Certificate is up to date and has not expired
    observedGeneration: 3
    reason: Ready
    status: "True"
    type: Ready
  notAfter: "2022-07-26T08:51:28Z"
  notBefore: "2021-07-26T08:51:28Z"
  renewalTime: "2022-03-26T16:51:28Z"
  revision: 2
```

That way we could also say the estimated time when the report will be deleted.
The "bad" thing is that users won't be able to change the TTL on the report, but I don't think I see this as a bad thing overall.
Another thing with the status field is that we could also show the TTL value when performing a get on the vulnerabilityreport and show another field. Don't know if we can do that with a feature flag or not though.

```shell
kubectl get vulnerabilityreports.aquasecurity.github.io
NAME                  REPOSITORY                       TAG       SCANNER   AGE          TTL
pod1                       thanos/thano                       v0.24.0      Trivy          14d            24h
```

This should probably be discussed in the design document. I will link back to this comment to make it visible.

danielpacak: VulnerabilityReport (and other reports managed by Starboard) are a bit different than Certificate and similar kinds. Mainly because we use them as data holder. Therefore the root property of the VulnerabilityReport is Report (I wish I have called it Data) and we do not have Spec. According to K8s API machinery conventions you either use Spec+Status or a single property with an arbitrary name (e.g. Report or Data).

danielpacak: Annotations are good to evaluate ideas without introducing new properties. Once we add a property it's hard to deprecate it. So I'd suggest to merge this PR with the annotation. We could always improve it if it's not enough.

For a helper method and parsing the annotation we could try defining it on the VulnerabilityReport type. Then whether we read it from annotation (today) or property (in few weeks or months) will be an implementation detail.

danielpacak: Anyway, I think in this particular line we inject TTL of type time.Duration, which is parsed by the env config module. If the value was incorrect the operator would not even start.

NissesSenap: Ahh didn't know that you couldn't use status if you used your own custom type. Thanks for the explanation

NissesSenap:
@danielpacak I have updated the builder now to use *time.duration and updated some docs and yaml to match the usage.

pkg/operator/controller/ttl_report_test.go
```diff
@@ -0,0 +1,28 @@
+package controller
+
+import (
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestTTLIsExpired(t *testing.T) {
```
danielpacak, line 10: Normally I would not buy these tests of a private method but it's okey for this merge request. A better practice for unit testing the TTL reconciler would be to test the `func (r *TTLReportReconciler) ReconcileReport() reconcile.Func` function, which is the only contract that we must obey.  The downside of testing private methods is that even if we don't change any core logic of the reconciler, changing just a private method would require rewriting unit tests. Such tests are not reliable and difficult to maintain. Some people even consider it an anti-pattern.

danielpacak (approved):
I think we are good to merge you excellent contribution @NissesSenap Thank you for all your efforts.

PR #890, opened 2022-01-07
refactor: rename kube.Object to kube.ObjectRef

danielpacak:
There is no good moment for such refactoring but I found
kube.Object pretty confusing with ubiquitous client.Object
defined by the controller-runtime module. We use both
types throughout the code in the same source files
so this patch renames kube.Object to kube.ObjectRef.

### DIFF
--- a/pkg/configauditreport/io.go
+++ b/pkg/configauditreport/io.go
@@ -28,18 +28,18 @@ type Writer interface {
 type Reader interface {
 
 	// FindReportByOwner returns a v1alpha1.ConfigAuditReport owned by the given
-	// kube.Object or nil if the report is not found.
-	FindReportByOwner(ctx context.Context, owner kube.Object) (*v1alpha1.ConfigAuditReport, error)
+	// kube.ObjectRef or nil if the report is not found.
+	FindReportByOwner(ctx context.Context, owner kube.ObjectRef) (*v1alpha1.ConfigAuditReport, error)
 
-	// FindReportByOwnerInHierarchy is similar to FindReportByOwner except that it tries to lookup
+	// FindReportByOwnerInHierarchy is similar to FindReportByOwner except that it tries to find
 	// a v1alpha1.ConfigAuditReport object owned by related Kubernetes objects.
 	// For example, if the given owner is a Deployment, but a report is owned by the
 	// active ReplicaSet (current revision) this method will return the report.
-	FindReportByOwnerInHierarchy(ctx context.Context, owner kube.Object) (*v1alpha1.ConfigAuditReport, error)
+	FindReportByOwnerInHierarchy(ctx context.Context, owner kube.ObjectRef) (*v1alpha1.ConfigAuditReport, error)
 
 	// FindClusterReportByOwner returns a v1alpha1.ClusterConfigAuditReport owned by the given
-	// kube.Object or nil if the report is not found.
-	FindClusterReportByOwner(ctx context.Context, owner kube.Object) (*v1alpha1.ClusterConfigAuditReport, error)
+	// kube.ObjectRef or nil if the report is not found.
+	FindClusterReportByOwner(ctx context.Context, owner kube.ObjectRef) (*v1alpha1.ClusterConfigAuditReport, error)
 }
 
 type ReadWriter interface {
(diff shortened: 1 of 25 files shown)


pkg/vulnerabilityreport/builder_test.go
```diff
@@ -99,7 +99,7 @@ func TestScanJobBuilder(t *testing.T) {
 	g.Expect(job).ToNot(gomega.BeNil())
 	g.Expect(job).To(gomega.Equal(&batchv1.Job{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      "scan-vulnerabilityreport-fcc9884cb",
+			Name:      "scan-vulnerabilityreport-64d65c457",
```
NissesSenap, line 102: No reason to update the test name that I can see.

pkg/plugin/conftest/plugin_test.go
```diff
@@ -402,7 +402,7 @@ func TestPlugin_GetScanJobSpec(t *testing.T) {
 		"Affinity":                     Equal(starboard.LinuxNodeAffinity()),
 		"Volumes": ConsistOf(
 			MatchFields(IgnoreExtras, Fields{
-				"Name": Equal("scan-configauditreport-5d4445db4f-volume"),
+				"Name": Equal("scan-configauditreport-789cbb5cc4-volume"),
```
NissesSenap, line 405: No reason to update the test name that I can see.

NissesSenap (commented):
I can only agree, nice to get it done.
I think you buy mistake have updated two test files. Other then that I think it looks good.

pkg/vulnerabilityreport/builder_test.go
danielpacak: It's the name of a scan job (not a test case) that has changed because we changed struct from `kube.Object` to `kube.ObjectRef` that is used to calculate the hash.

pkg/plugin/conftest/plugin_test.go
danielpacak: Also here, the Name property is the name of a scan job not a test case.

pkg/configauditreport/io.go
```diff
@@ -29,17 +29,17 @@ type Reader interface {
 
 	// FindReportByOwner returns a v1alpha1.ConfigAuditReport owned by the given
 	// kube.Object or nil if the report is not found.
```
deven0t, line 31: I think you missed updating the comment part here. It is still referring to kube.Object

pkg/configauditreport/io.go
```diff
@@ -29,17 +29,17 @@ type Reader interface {
...
 
 	// FindReportByOwnerInHierarchy is similar to FindReportByOwner except that it tries to lookup
 	// a v1alpha1.ConfigAuditReport object owned by related Kubernetes objects.
 	// For example, if the given owner is a Deployment, but a report is owned by the
 	// active ReplicaSet (current revision) this method will return the report.
-	FindReportByOwnerInHierarchy(ctx context.Context, owner kube.Object) (*v1alpha1.ConfigAuditReport, error)
+	FindReportByOwnerInHierarchy(ctx context.Context, owner kube.ObjectRef) (*v1alpha1.ConfigAuditReport, error)
 
 	// FindClusterReportByOwner returns a v1alpha1.ClusterConfigAuditReport owned by the given
 	// kube.Object or nil if the report is not found.
```
deven0t, line 41: same here, the comment is referring to kube.Object.

pkg/vulnerabilityreport/io.go
```diff
@@ -29,8 +29,8 @@ type Writer interface {
 // For example, if the given owner is a Deployment, but reports are owned by the
```
deven0t, line 29: same here, comment for FindByOwner is referring to kube.Object

deven0t (commented):
Yes, agree with refactoring change. changes LGTM. 
Few places kube.Object reference is still there, this is in the comment section of code. I think we can change them to kube.ObjectRef.

pkg/configauditreport/io.go
danielpacak: Good catch! Thanks. I've updated Go docs as well.

deven0t (approved):

PR #917, opened 2022-01-24
docs: add design doc for schedule vuln scan job in workload's namespace

deven0t:
resolves: #817


docs/design/design_vuln_scan_job_in_same_namespace_of_workload.md
```diff
@@ -0,0 +1,81 @@
...
+Currently, when image is pulled from cloud registries(ecr, acr) in cluster and authentication is set on service account. Then 
+Trivy plugin cannot use that authentication to pull cloud registry image. If user want to get those images scanned then they
+have to specify those credentials at starboard service account([ECR registry configuration]). Also, recently there was one
+option added in Trivy plugin to use Trivy filesystem scan, in which starboard doesnt have option to reuse ImagePullSecret 
+defined on workload spec or on service account. To resolve this limitation, if starboard manages to run scan job in same 
+namespace where workload is running then we can reuse the service account and imagePullSecret available with the workload.
+
+## Solution
+
+Add an environment variable to operator eg `OPERATOR_ENABLE_SCAN_JOB_IN_SAME_NAMESPACE=true`. This way operator can schedule
```
chen-keinan, line 14: Looks Great !!
Couple of comments: 
- The parameter name is generic (for all jobs), however it applies for `vulnerability scanning` job only, for now, just to make sure this reflected in documentation
- looking forward in the future if we want this env. Var to support other jobs types (kube-bench / confaudit and etc) --> since this job will inherit the service account of the workload, it might have some limitation for other jobs which are not `Scan vulnerability` as different jobs might require different permission

deven0t: thanks Chen, yes agree with you that this option will have some limitations for other scans considering the permission barrier. I will change it to `OPERATOR_ENABLE_VULNERABILITY_SCAN_JOB_IN_SAME_NAMESPACE`.

danielpacak: Notice that params related to vulnerability scanners are prefixed with `OPERATOR_VULNERABILITY_SCANNER_`

OPERATOR_VULNERABILITY_SCANNER_ENABLED
OPERATOR_VULNERABILITY_SCANNER_SCAN_ONLY_CURRENT_REVISIONS
OPERATOR_VULNERABILITY_SCANNER_REPORT_TTL

docs/design/design_vuln_scan_job_in_same_namespace_of_workload.md
```diff
@@ -0,0 +1,81 @@
+# Run Vulnerability Scan Job in Same namespace of workload
+
+## Overview
```
danielpacak, line 3: Running scan jobs in the same namespace as workload is an implementation detail or a prerequisite to scan particular workloads that use images from private managed registries.

What we should clarify is why we even consider running scan jobs in the same namespace. In particular, we should explain how ECR injects credentials as extra environments variables passed to containers that run with authorized service account. IIRC in EKS integrated with ECR there is this mutating webhook that does the trick. So we want to run scan jobs in the same namespace because we want to use workload's service account instead of the `starboard-operator` service account. By doing that the mutating webhooks will add some extra variables to scan containers (created by Starboard) that are eventually passed to Trivy. Finally, Trivy can recognise injected AWS environment variables and use to pull down private image from managed registry. Thus, imagePullSecrets are not required.

Another thing I'd like us to clarify is Trivy FS scanner and why we consider it along with same namespace config / design proposal? I remember we discussed it that if we pass image pull secrets to Trivy FS scanner it somehow falls back to the current behaviour, where Trivy pull image from the remote registry and is not scanning image cached on K8s node.

docs/design/design_vuln_scan_job_in_same_namespace_of_workload.md
```diff
@@ -0,0 +1,81 @@
...
+
+## Solution
+
+Add an environment variable to operator eg `OPERATOR_ENABLE_SCAN_JOB_IN_SAME_NAMESPACE=true`. This way operator can schedule
+and monitor scan jobs in same namespace where workload is running. And plugins will act accordingly to utilize the service 
+account and ImagePullSecret available on the workload.
+
+> :bulb: For this proposal this option will be implemented only for vulnerability scans.
+
+### Example
```
danielpacak, line 20: Can we add a realistic example base on ECR and IAM where we don't really have imagePullSecret but there's a service account that's annotated in some way to grant it image pull permissions.

Another example might be Trivy FS scanner with imagePullSecret if we consider it.

danielpacak (changes requested):
👋 @deven0t I think it's a good start. Please see my request for adding a bit more context and reasons why we even consider creating scan jobs in the same namespace as scanned workload.

I think it's also worth mentioning some cons of this solution in comparison to creating everything in the `starboard-system` namespace.

docs/design/design_vuln_scan_job_in_same_namespace_of_workload.md
deven0t: thanks @danielpacak , I have added some details to explain the motivation for this proposal. Please review it again.

danielpacak (approved):

chen-keinan (approved):

PR #934, opened 2022-01-27
feat/nsa plugin support

chen-keinan:
This is an initial draft POC.
What is does?
1. introduce new NSA controller which encapsulate config-audit and cis-benchmark controller.
2. listen to resource event and activate the right tool (generate NSA scan jobs) based on resource kind
 - cis-benchmark for `Node` kind
 - configAudit for `resource` kind
 3. Create NSA reports from the output of the scan jobs and publish to Etcd

Note: further more improvements are required to be made to make it robust 
          test need to be fixed and added


chen-keinan:
Closing PR we decided to go with different approch

PR #943, opened 2022-02-01
feat: Add option to run Vuln scan job in worload namespace

deven0t:
new option in configmap starboard has added ie. ``vulnerabilityReports.scanJobsInSameNamespace`` to enable running
scan job in the workload namespace.
Once this option is enabled then vulnerability scan jobs will start running in same namespace of workload.
Also it will run with same serviceAccount and imagepullSecret of workload.
This option helps us to schedule scan job with Trivy fs command on any node.

this implements the approach mentioned here: https://github.com/aquasecurity/starboard/pull/917

### DIFF
--- a/pkg/operator/controller/limit_checker.go
+++ b/pkg/operator/controller/limit_checker.go
@@ -13,16 +13,18 @@ type LimitChecker interface {
 	Check(ctx context.Context) (bool, int, error)
 }
 
-func NewLimitChecker(config etc.Config, client client.Client) LimitChecker {
+func NewLimitChecker(config etc.Config, client client.Client, starboardConfig starboard.ConfigData) LimitChecker {
 	return &checker{
-		config: config,
-		client: client,
+		config:          config,
+		client:          client,
+		starboardConfig: starboardConfig,
 	}
 }
 
 type checker struct {
-	config etc.Config
-	client client.Client
+	config          etc.Config
+	client          client.Client
+	starboardConfig starboard.ConfigData
 }
 
 func (c *checker) Check(ctx context.Context) (bool, int, error) {
@@ -36,9 +38,14 @@ func (c *checker) Check(ctx context.Context) (bool, int, error) {
 
 func (c *checker) countScanJobs(ctx context.Context) (int, error) {
 	var scanJobs batchv1.JobList
-	err := c.client.List(ctx, &scanJobs, client.MatchingLabels{
+	listOptions := []client.ListOption{client.MatchingLabels{
 		starboard.LabelK8SAppManagedBy: starboard.AppStarboard,
-	}, client.InNamespace(c.config.Namespace))
+	}}
+	if !c.starboardConfig.VulnerabilityScanJobsInSameNamespace() {
+		// scan jobs are running in only starboard operator namespace
+		listOptions = append(listOptions, client.InNamespace(c.config.Namespace))
+	}
+	err := c.client.List(ctx, &scanJobs, listOptions...)
 	if err != nil {
 		return 0, err
 	}
--- a/pkg/operator/controller/vulnerabilityreport.go
+++ b/pkg/operator/controller/vulnerabilityreport.go
@@ -23,6 +23,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
@@ -73,13 +74,13 @@ func (r *VulnerabilityReportReconciler) SetupWithManager(mgr ctrl.Manager) error
 			return err
 		}
 	}
+	var predicates []predicate.Predicate
+	if !r.ConfigData.VulnerabilityScanJobsInSameNamespace() {
+		predicates = append(predicates, InNamespace(r.Config.Namespace))
+	}
+	predicates = append(predicates, ManagedByStarboardOperator, IsVulnerabilityReportScan, JobHasAnyCondition)
 	return ctrl.NewControllerManagedBy(mgr).
-		For(&batchv1.Job{}, builder.WithPredicates(
-			InNamespace(r.Config.Namespace),
-			ManagedByStarboardOperator,
-			IsVulnerabilityReportScan,
-			JobHasAnyCondition,
-		)).
+		For(&batchv1.Job{}, builder.WithPredicates(predicates...)).
 		Complete(r.reconcileJobs())
 }
 
@@ -260,7 +261,6 @@ func (r *VulnerabilityReportReconciler) submitScanJob(ctx context.Context, owner
 	}
 
 	for _, secret := range secrets {
-		secret.Namespace = r.PluginContext.GetNamespace()
 		err = r.Client.Create(ctx, secret)
 		if err != nil {
 			if k8sapierror.IsAlreadyExists(err) {
--- a/pkg/plugin/factory.go
+++ b/pkg/plugin/factory.go
@@ -75,6 +75,7 @@ func (r *Resolver) GetVulnerabilityPlugin() (vulnerabilityreport.Plugin, starboa
 		WithNamespace(r.namespace).
 		WithServiceAccountName(r.serviceAccountName).
 		WithClient(r.client).
+		WithStarboardConfig(r.config).
 		Get()
 
 	switch scanner {
--- a/pkg/plugin/trivy/plugin.go
+++ b/pkg/plugin/trivy/plugin.go
@@ -956,10 +956,17 @@ func (p *plugin) getPodSpecForStandaloneFSMode(ctx starboard.PluginContext, conf
 	if err != nil {
 		return corev1.PodSpec{}, nil, err
 	}
-	// get nodeName from running pods.
-	spec.NodeName, err = p.objectResolver.GetNodeName(context.Background(), workload)
-	if err != nil {
-		return corev1.PodSpec{}, nil, err
+	pullPolicy := corev1.PullIfNotPresent
+	// nodeName to schedule scan job explicitly on specific node.
+	var nodeName string
+	if !ctx.GetStarboardConfig().VulnerabilityScanJobsInSameNamespace() {
+		// get nodeName from running pods.
+		nodeName, err = p.objectResolver.GetNodeName(context.Background(), workload)
+		if err != nil {
+			return corev1.PodSpec{}, nil, fmt.Errorf("failed resolving node name for workload %q: %w",
+				workload.GetNamespace()+"/"+workload.GetName(), err)
+		}
+		pullPolicy = corev1.PullNever
 	}
 
 	trivyImageRef, err := config.GetImageRef()
@@ -1103,7 +1110,7 @@ func (p *plugin) getPodSpecForStandaloneFSMode(ctx starboard.PluginContext, conf
 		containers = append(containers, corev1.Container{
 			Name:                     c.Name,
 			Image:                    c.Image,
-			ImagePullPolicy:          corev1.PullNever,
+			ImagePullPolicy:          pullPolicy,
 			TerminationMessagePolicy: corev1.TerminationMessageFallbackToLogsOnError,
 			Env:                      env,
 			Command: []string{
@@ -1135,7 +1142,7 @@ func (p *plugin) getPodSpecForStandaloneFSMode(ctx starboard.PluginContext, conf
 		})
 	}
 
-	return corev1.PodSpec{
+	podSpec := corev1.PodSpec{
 		Affinity:                     starboard.LinuxNodeAffinity(),
 		RestartPolicy:                corev1.RestartPolicyNever,
 		ServiceAccountName:           ctx.GetServiceAccountName(),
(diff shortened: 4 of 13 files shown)


docs/operator/configuration.md
```diff
@@ -17,6 +17,7 @@ Configuration of the operator's Pod is done via environment variables at startup
 | `OPERATOR_VULNERABILITY_SCANNER_ENABLED`                     | `true`               | The flag to enable vulnerability scanner                                                                                                                                                                     |
 | `OPERATOR_CONFIG_AUDIT_SCANNER_ENABLED`                      | `true`               | The flag to enable configuration audit scanner                                                                                                                                                               |
 | `OPERATOR_VULNERABILITY_SCANNER_SCAN_ONLY_CURRENT_REVISIONS` | `false`              | The flag to enable vulnerability scanner to only scan the current revision of a deployment                                                                                                                   |
+| `OPERATOR_ENABLE_SCAN_JOB_IN_SAME_NAMESPACE`                 | `false`              | The flag to enable running scan jobs in the same namespace of scanned resource. Currently this option is enabled for only vulnerability scanning                                                             |
```
danielpacak, line 20: Use common prefix for vulnerability scanner configs `OPERATOR_VULNERABILITY_SCANNER_`

pkg/operator/controller/limit_checker.go
```diff
@@ -36,9 +36,14 @@ func (c *checker) Check(ctx context.Context) (bool, int, error) {
 
 func (c *checker) countScanJobs(ctx context.Context) (int, error) {
 	var scanJobs batchv1.JobList
-	err := c.client.List(ctx, &scanJobs, client.MatchingLabels{
+	listOptions := []client.ListOption{client.MatchingLabels{
 		starboard.LabelK8SAppManagedBy: starboard.AppStarboard,
-	}, client.InNamespace(c.config.Namespace))
+	}}
+	if !c.config.ScanJobInSameNamespace {
```
danielpacak, line 42: The same instance of LimitChecker is use by other controllers, whereas `config.ScanJobInSameNamespace` is only pertinent to vulnerability scans. (We should probably rename it to `config.VulnerabilityScannerSameNamespace` to make it more explicit.)

One solution might be to create another instance of LimitCheck that's aware of such config and inject it into VulnerabilityReportReconciler.

pkg/operator/etc/config.go
```diff
@@ -25,6 +25,7 @@ type Config struct {
 	VulnerabilityScannerEnabled                  bool           `env:"OPERATOR_VULNERABILITY_SCANNER_ENABLED" envDefault:"true"`
 	VulnerabilityScannerScanOnlyCurrentRevisions bool           `env:"OPERATOR_VULNERABILITY_SCANNER_SCAN_ONLY_CURRENT_REVISIONS" envDefault:"false"`
 	VulnerabilityScannerReportTTL                *time.Duration `env:"OPERATOR_VULNERABILITY_SCANNER_REPORT_TTL"`
+	ScanJobInSameNamespace                       bool           `env:"OPERATOR_ENABLE_SCAN_JOB_IN_SAME_NAMESPACE" envDefault:"false"`
```
danielpacak, line 28: See other params related to vulnerability scanning:
```suggestion
	VulnerabilityScannerSameNamespace                       bool           `env:"OPERATOR_VULNERABILITY_SCANNER_SAME_NAMESPACE" envDefault:"false"`
```

pkg/plugin/factory.go
```diff
@@ -20,6 +21,7 @@ type Resolver struct {
 	namespace          string
 	serviceAccountName string
 	client             client.Client
+	operatorConfig     etc.Config
```
danielpacak, line 24: PluginFactory is used by CLI and Operator. So I'm wondering if instead of configuring this behaviour with environment variable it should to stored in the `starboard` ConfigMap that is ensured in the starboard-system namespace. WDYT?

By injecting etc.Config we implicitly make CLI aware of Go packages designed for Operator.

pkg/starboard/plugin.go
```diff
@@ -44,6 +46,9 @@ type PluginContext interface {
 	// GetServiceAccountName return the name of the K8s Service Account used to run workloads
 	// created by Starboard.
 	GetServiceAccountName() string
+	// GetOperatorConfig return the starboard operator config instance, with which starboard operator
+	// is running.
+	GetOperatorConfig() etc.Config
```
danielpacak, line 51: See my comment related to PluginFactory and etc.Config. I think instead we could pass (generic) Starboard config to the PluginContext.

danielpacak (changes requested):
Hey @deven0t Good progress! I've run the code in my cluster, did a basic test, and it seems to be working fine. We still need to test it in different multi tenancy modes.

Anyway, see my comments regarding operator config and (generic) starboard settings and why I think we may reconsider `OPERATOR_VULNERABITLY_SCANNER_SAME_NAMESPACE` environment variable and introduce `vulnerabiliyReports.scanJobsInSameNamespace` [settings](https://aquasecurity.github.io/starboard/v0.14.1/settings/) param.

chen-keinan:
@deven0t this flow of vulnerability scan in workload namespace is enabled via config parameter, which goes across the flow of creating a vulnerability scan job, is there a way to add a test which validates this flow, so we could protect it ?

pkg/operator/controller/limit_checker.go
deven0t: As discussed, this checker will check in entire cluster when `samenamespace` option is enabled. So I think this code should work fine.

deven0t:
> @deven0t this flow of vulnerability scan in workload namespace is enabled via config parameter, which goes across the flow of creating a vulnerability scan job, is there a way to add a test which validates this flow, so we could protect it ?

Thanks @chen-keinan for feedback, after discussion with Danial we have decided to change that configuration from env to starboard configmap. I have added some unit test cases to validate this flow. Please review it again.

chen-keinan:
@deven0t great job!! LGTM

deven0t:
Hi @danielpacak @chen-keinan, I think we missed one point in this proposal and in implementation as well. How plugin config will be available for scan job in the namespace of workload. eg trivy.severity, trivy.githubToken. Currently, the scan job doesn't fail with errors coz these are optional ENV vars. But if we don't provide them in the namespace of workload then the scan job will run without these configs.

danielpacak:
> Hi @danielpacak @chen-keinan, I think we missed one point in this proposal and in implementation as well. How plugin config will be available for scan job in the namespace of workload. eg trivy.severity, trivy.githubToken. Currently, the scan job doesn't fail with errors coz these are optional ENV vars. But if we don't provide them in the namespace of workload then the scan job will run without these configs.

That's a very good catch @deven0t We resolved this problem in config audits by copying a snapshot of plugin config to a temporary Secret that's owned by a scan job. And I think we should do the same for Trivy plugin. In general, someone might update starboard-trivy-config by the time the scan pod is actually run and we'll bump into this problem even if we run scan jobs in the starboard-system namespace.

danielpacak:
Hey @deven0t As we discussed I think we should mark this PR as draft and implement similar mechanism to store Trivy config snapshot in a temporary Secret instead of using the `starboard-trivy-config` ConfigMap directly.

deven0t:
> Hey @deven0t As we discussed I think we should mark this PR as draft and implement similar mechanism to store Trivy config snapshot in a temporary Secret instead of using the `starboard-trivy-config` ConfigMap directly.

Hi @danielpacak , as we discussed this change will not be enabled by default, so it will not impact any existing flow. We can proceed with this PR and later with the new PR we will add a mechanism to create temp secret and use it for scan job purpose.

pkg/vulnerabilityreport/builder.go
```diff
@@ -129,6 +129,11 @@ func (s *ScanJobBuilder) Get() (*batchv1.Job, []*corev1.Secret, error) {
 			},
 		},
 	}
+	// secrets will be created with scan jobs in same namespace where scan job will run
+	for i, _ := range secrets {
+		secrets[i].Namespace = s.pluginContext.GetNamespace()
+	}
```
danielpacak, line 135: Isn't this code duplicated within the updateScanJobForWorkloadNamespace method?

pkg/vulnerabilityreport/builder_test.go
```diff
@@ -64,76 +64,152 @@ func TestReportBuilder(t *testing.T) {
...
+
+	t.Run("Should get scan job running in workload namespace", func(t *testing.T) {
+		g := gomega.NewGomegaWithT(t)
+		job, _, err := vulnerabilityreport.NewScanJobBuilder().
+			WithPlugin(&testPlugin{}).
+			WithPluginContext(starboard.NewPluginContext().
+				WithName("test-plugin").
+				WithNamespace("starboard-ns").
+				WithServiceAccountName("starboard-sa").
+				WithStarboardConfig(map[string]string{starboard.KeyVulnerabilityScansInSameNamespace: "true"}).
```
danielpacak, line 148:
```suggestion
				WithStarboardConfig(starboard.ConfigData{
					starboard.KeyVulnerabilityScansInSameNamespace: "true",
				}).
```

pkg/plugin/trivy/plugin.go
```diff
@@ -1134,8 +1136,12 @@ func (p *plugin) getPodSpecForStandaloneFSMode(ctx starboard.PluginContext, conf
 			},
 		})
 	}
-
-	return corev1.PodSpec{
+	if ctx.GetStarboardConfig().IsScanJobsInSameNamespace() {
+		for i, _ := range containers {
```
danielpacak, line 1140: Do we need this loop? Maybe we can determine image pull policy in the beginning of the method and just set it when we create containers?

pkg/plugin/trivy/plugin.go
```diff
@@ -956,10 +956,12 @@ func (p *plugin) getPodSpecForStandaloneFSMode(ctx starboard.PluginContext, conf
...
 	}
-	// get nodeName from running pods.
-	spec.NodeName, err = p.objectResolver.GetNodeName(context.Background(), workload)
-	if err != nil {
-		return corev1.PodSpec{}, nil, err
+	if !ctx.GetStarboardConfig().IsScanJobsInSameNamespace() {
+		// get nodeName from running pods.
+		spec.NodeName, err = p.objectResolver.GetNodeName(context.Background(), workload)
+		if err != nil {
+			return corev1.PodSpec{}, nil, err
```
danielpacak, line 963: Let's add more context in case of failure so we will see exactly in the logs what is going on and which workload is problematic:
```suggestion
			return corev1.PodSpec{}, nil, fmt.Errorf("failed resolving node name for workload %q: %w", workload.Namespace+"/"+workload.Name, err)
```

pkg/plugin/trivy/plugin.go
```diff
@@ -1144,8 +1150,13 @@ func (p *plugin) getPodSpecForStandaloneFSMode(ctx starboard.PluginContext, conf
 		InitContainers:               []corev1.Container{initContainerCopyBinary, initContainerDB},
 		Containers:                   containers,
 		SecurityContext:              &corev1.PodSecurityContext{},
-		NodeName:                     spec.NodeName,
-	}, secrets, nil
+	}
+
+	if !ctx.GetStarboardConfig().IsScanJobsInSameNamespace() {
+		podSpec.NodeName = spec.NodeName
```
danielpacak, line 1156: I see that we take the value form spec.NodeName based on p.objectResolver.GetNodeName(). But why don't we use objectResolver directly here?

danielpacak (commented):
👋 @deven0t I left a few minor comments and questions to make sure I'm following the changes

pkg/operator/controller/vulnerabilityreport.go
```diff
@@ -73,13 +74,13 @@ func (r *VulnerabilityReportReconciler) SetupWithManager(mgr ctrl.Manager) error
 			return err
 		}
 	}
+	var predicates []predicate.Predicate
+	if !r.ConfigData.IsScanJobsInSameNamespace() {
+		predicates = append(predicates, InNamespace(r.Config.Namespace))
+	}
+	predicates = append(predicates, ManagedByStarboardOperator, IsVulnerabilityReportScan, JobHasAnyCondition)
```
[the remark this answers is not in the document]
deven0t: reason for doing other way round, so that InNamespace predicate will get hit before other predicates. If the sequence of the predicates is not guaranteed then we can go ahead with your suggested way.

pkg/vulnerabilityreport/builder.go
deven0t: This is existing code that was there in [pkg/operator/controller/vulnerabilityreport.go](https://github.com/aquasecurity/starboard/pull/943/files#diff-6e57a58bb4c7202d8e6d177078a99e7d95b3e436dd3f4e0e90eb990a1e2e4987L262) Due to that code update in updateScanJobForWorkloadNamespace method was not reflecting at the time of secret creation.

So this namespace set is required regardless of sameNamespace option, so that secrets are created in starboad-operator ns. We can do this inside plugin, but since builder is setting namespace for job, so setting it here.

pkg/plugin/trivy/plugin.go
deven0t: GetNodeName() returns error as well and when there is error then we dont want to proceed for building other items(initcontainer, vol, etc). So get nodeName at the start and using it here. wdyt

danielpacak: We can call it as early as possible, but my point is that adding a dedicated nodeName variable is more readable then updating spec.

deven0t: ohh, I got the point. thanks.

danielpacak (approved):

PR #950, opened 2022-02-06
fix: ignore parsing logs of deleted scan jobs

### LINKED ISSUE #873: Ignore parsing logs of scan jobs have been deleted
In the configauditreports controller we should ignore scan jobs that are cached but had been deleted from etcd so we cannot parse their logs. Currently we return an error, which is causing retry, but retrying does not make much sense in this case.

https://github.com/aquasecurity/starboard/blob/v0.13.0/pkg/operator/controller/configauditreport.go#L406

mycodeself:
fixes #873 based on @py-go suggestion

Avoid retry parse deleted scan jobs handling the `IsNotFound` error and deleting the job.

### DIFF
--- a/pkg/operator/controller/configauditreport.go
+++ b/pkg/operator/controller/configauditreport.go
@@ -409,7 +409,12 @@ func (r *ConfigAuditReportReconciler) processCompleteScanJob(ctx context.Context
 
 	logsStream, err := r.LogsReader.GetLogsByJobAndContainerName(ctx, job, r.Plugin.GetContainerName())
 	if err != nil {
-		return fmt.Errorf("getting logs: %w", err)
+		if errors.IsNotFound(err) {
+			log.V(1).Info("Ignoring not found job", "job", fmt.Sprintf("%s/%s", job.Namespace, job.Name), "owner", owner)
+			log.V(1).Info("Deleting complete scan job", "owner", owner)
+			return r.deleteJob(ctx, job)
+		}
+		return fmt.Errorf("getting pod controlled by job: %q: %w", job.Namespace+"/"+job.Name, err)
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/operator/controller/configauditreport.go
```diff
@@ -409,7 +409,12 @@ func (r *ConfigAuditReportReconciler) processCompleteScanJob(ctx context.Context
 
 	logsStream, err := r.LogsReader.GetLogsByJobAndContainerName(ctx, job, r.Plugin.GetContainerName())
 	if err != nil {
-		return fmt.Errorf("getting logs: %w", err)
+		if errors.IsNotFound(err) {
+			log.V(1).Info("Ignoring not found job", "job", fmt.Sprintf("%s/%s", job.Namespace, job.Name), "owner", owner)
+			log.V(1).Info("Deleting complete scan job", "owner", owner)
+			return r.deleteJob(ctx, job)
```
chen-keinan, line 415: @mycodeself Probably `r.deleteJob` will not find a job to delete, due to job `not found`.

Maybe is better just to `return ctrl.Result{}, nil` to remove it from the cache :
https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/internal/controller/controller.go#L332

mycodeself: @chen-keinan but `return ctrl.Result{}, nil` doesn't fit the signature of `processCompleteScanJob` it just returns an error. I guess that I should just `return nil` right?

chen-keinan: @mycodeself FYI, The complete solution has been implemented by #956

mycodeself: Oh ok @chen-keinan thanks for letting me know! I'll watch this PR to continue learning!

danielpacak:
Let's close this one and keep working on #956.

PR #956, opened 2022-02-09
Fix: Deletes scan jobs if their pods have been deleted and logs can't be c…

### LINKED ISSUE #873: Ignore parsing logs of scan jobs have been deleted

jw-s:
Fixes #873 and for other scan types which are not mentioned in this issue

The problem which is fixed in this PR is the following:

When a scan job creates the pod which handles scans, if this pod has been deleted before starboard parses the pod's logs, these jobs are never deleted which means when the max concurrent job setting is hit, part of this threshold is eaten up by jobs which never get deleted.

Signed-off-by: Joel Whittaker-Smith <jdws.dev@gmail.com>

### DIFF
--- a/pkg/kube/logs.go
+++ b/pkg/kube/logs.go
@@ -2,6 +2,7 @@ package kube
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"io"
 
@@ -10,6 +11,8 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
 )
+
+var podControlledByJobNotFoundErr = errors.New("pod for job not found")
 
 type LogsReader interface {
 	GetLogsByJobAndContainerName(ctx context.Context, job *batchv1.Job, containerName string) (io.ReadCloser, error)
@@ -32,7 +35,7 @@ func (r *logsReader) GetLogsByJobAndContainerName(ctx context.Context, job *batc
 		return nil, fmt.Errorf("getting pod controlled by job: %q: %w", job.Namespace+"/"+job.Name, err)
 	}
 	if pod == nil {
-		return nil, fmt.Errorf("getting pod controlled by job: %q: pod not found", job.Namespace+"/"+job.Name)
+		return nil, fmt.Errorf("getting pod controlled by job: %q: %w", job.Namespace+"/"+job.Name, podControlledByJobNotFoundErr)
 	}
 
 	return r.clientset.CoreV1().Pods(pod.Namespace).
@@ -86,4 +89,8 @@ func GetTerminatedContainersStatusesByPod(pod *corev1.Pod) map[string]*corev1.Co
 		states[status.Name] = status.State.Terminated
 	}
 	return states
+}
+
+func IsPodControlledByJobNotFound(err error) bool {
+	return errors.Is(err, podControlledByJobNotFoundErr)
 }
--- a/itest/starboard-operator/suite_test.go
+++ b/itest/starboard-operator/suite_test.go
@@ -66,9 +66,8 @@ var _ = BeforeSuite(func() {
 		AssertTimeout:         3 * time.Minute,
 		PrimaryNamespace:      corev1.NamespaceDefault,
 		PrimaryWorkloadPrefix: "wordpress",
-
-		Client: kubeClient,
-		Helper: helper.NewHelper(kubeClient),
+		Client:                kubeClient,
+		Helper:                helper.NewHelper(kubeClient),
 	}
 
 	startCtx, stopFunc = context.WithCancel(context.Background())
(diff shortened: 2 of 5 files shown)


pkg/kube/logs.go
```diff
@@ -32,7 +35,7 @@ func (r *logsReader) GetLogsByJobAndContainerName(ctx context.Context, job *batc
 		return nil, fmt.Errorf("getting pod controlled by job: %q: %w", job.Namespace+"/"+job.Name, err)
 	}
 	if pod == nil {
```
chen-keinan, line 37: @jw-s thank you for the contribution:
maybe we should handle both cases :
1. Job deleted from Etcd, but the job exist in controller cache
2. pod deleted from Etcd and Job exist in Etcd and cache (as appear in your PR)

jw-s: Hi @chen-keinan Point 1 is covered before it even gets to this stage, earlier on in the reconcile look it checks if the object exists for example: https://github.com/aquasecurity/starboard/blob/main/pkg/operator/controller/vulnerabilityreport.go#L97

chen-keinan: @jw-s to `reconcile loop` you mention is for the `workload`, I'm referring to the 2nd reconcile loop for `scan jobs`
where 1st we tried to get it from cache :
`log.V(1).Info("Getting job from cache")
		err := r.Client.Get(ctx, req.NamespacedName, job)
		if err != nil {
                                if errors.IsNotFound(err) {
				log.V(1).Info("Ignoring cached job that must have been deleted")
				return ctrl.Result{}, nil
			}`
and only afterwards we get tries to get the job from Etcd :
`
func (r *logsReader) getPodByJob(ctx context.Context, job *batchv1.Job) (*corev1.Pod, error) {
	refreshedJob, err := r.clientset.BatchV1().Jobs(job.Namespace).Get(ctx, job.Name, metav1.GetOptions{})
	`

chen-keinan: @jw-s It will be great if you could add the check for `Job not found`,  referring to other PR opened on same issue #950

jw-s: @chen-keinan The new commit checks if the job isn't found and returns nil causing the controller to no requeue.

danielpacak (changes requested):
Thank you for working on this enhancement @jw-s Overall the controllers code looks good to me. 

The only thing I'm not comfortable with are integration tests that you wrote for "relatively rare" corner case when the operator's cache cannot catch up with K8s API Server state and parse logs.

What's more, I'm afraid that the integration tests became flaky because we cannot reliably control Pods managed by JobController, which is part of K8s controller-manager. (See failed PR checks.)

To sum up, I'd just add unit tests whenever possible (and valuable). Let's keep integration (~ e2e) test nice and lean so we can afford running them on each PR.

jw-s:
> Thank you for working on this enhancement @jw-s Overall the controllers code looks good to me.
> 
> The only thing I'm not comfortable with are integration tests that you wrote for "relatively rare" corner case when the operator's cache cannot catch up with K8s API Server state and parse logs.
> 
> What's more, I'm afraid that the integration tests might became flaky because we cannot reliably control Pods managed by JobController, which is part of K8s controller-manager. (See failed PR checks.)
> 
> To sum up, I'd just add unit tests whenever possible to keep integration (~ e2e) test nice and lean so we can afford running them on each PR. See this nice and detailed explanation why I think we should not validate everything at the integration tests phase [martinfowler.com/articles/practical-test-pyramid.html](https://martinfowler.com/articles/practical-test-pyramid.html).

Yeah I completely agree, this area is definitely flake provoking and i'm happy to remove these tests. Do we have an area where we can add unit tests to cover these cases? I couldn't seem to find it.

danielpacak:
> > Thank you for working on this enhancement @jw-s Overall the controllers code looks good to me.
> > The only thing I'm not comfortable with are integration tests that you wrote for "relatively rare" corner case when the operator's cache cannot catch up with K8s API Server state and parse logs.
> > What's more, I'm afraid that the integration tests might became flaky because we cannot reliably control Pods managed by JobController, which is part of K8s controller-manager. (See failed PR checks.)
> > To sum up, I'd just add unit tests whenever possible to keep integration (~ e2e) test nice and lean so we can afford running them on each PR. See this nice and detailed explanation why I think we should not validate everything at the integration tests phase [martinfowler.com/articles/practical-test-pyramid.html](https://martinfowler.com/articles/practical-test-pyramid.html).
> 
> Yeah I completely agree, this area is definitely flake provoking and i'm happy to remove these tests (Though the failure is not because of a flake but actually something else). Do we have an area where we can add unit tests to cover these cases? I couldn't seem to find it.

We'd have to add _test.go files to implement unit tests for reconcilers. So far we only have integration tests for controllers. If this requires too much effort I'm also happy to approve this PR without very strict requirement for unit testing.

danielpacak (approved):
Thank you @jw-s for the contribution. The code LGTM!

Issue #889, opened 2022-01-07
Revisit architecture of Starboard Operator and configuration auditing


danielpacak:
To audit configuration of built-in K8s objects Starboard Operator creates Jobs. Each Job runs the Conftest container to  evaluate OPA Rego policies stored as key-value pairs in the `starboard-conftest-config` ConfigMap.

Even though Starboard Operator keeps a limit of concurrent scan Jobs (`OPERATOR_CONCURRENT_SCAN_JOBS_LIMIT`) the current solution is suboptimal and might flood cluster with too many Job objects.

The task is to consider alternative approach to configuration audits based on OPA Rego policies stored as key-value pairs in the `starboard-conftest-config` ConfigMap.

![starboard-configaudit-jobs](https://user-images.githubusercontent.com/1322923/148553801-b10533a3-3f2a-4db4-97d4-f0710cc2d4ad.png)

PR #971, opened 2022-02-18
feat: built-in OPA Rego policy-based configuration audit scanner

danielpacak:
1. Perform synchronous configuration audit within reconciliation loop instead of
   creating Kubernetes Job objects and associated Secrets.
2. Configuration audits at scale with build-in controller-runtime workers.
3. Support NetworkPolicies, PodSecurityPolicies, Ingress, ResourceQuota, and
   possibly other resources.
4. Use OPA Go SDK to evaluate policies read from the `starboard-policy-config`
   ConfigMap, which is compatible with Conftest plugin and KubeEnforcer.
   This is just an MVP, but for production workloads OPA policies can be
   prepared for evaluation and cached similarly to what OPA server does.
5. Preinstall all Kubernetes policies from
   https://github.com/aquasecurity/defsec/tree/v0.14.0/rules/kubernetes
6. First step to deprecate and remove Polaris and Conftest plugins to streamline
   code, usages, and documentation.

Resolves: #889


pkg/policy/policy.go
```diff
@@ -0,0 +1,325 @@
...
+
+// NewMetadata constructs new Metadata based on raw value.
+func NewMetadata(value map[string]interface{}) (Metadata, error) {
+	severity, err := v1alpha1.StringToSeverity(fmt.Sprintf("%v", value["severity"]))
+	if err != nil {
+		return Metadata{}, fmt.Errorf("failed parsing severity from raw value: %w", err)
+	}
+	return Metadata{
+		Severity:    severity,
+		ID:          fmt.Sprintf("%v", value["id"]),
```
chen-keinan, line 45: value["id"] is not checked for existence , id can be set for "<nil>"

pkg/policy/policy.go
```diff
@@ -0,0 +1,325 @@
...
+	Title    string
+	Severity v1alpha1.Severity
+	Type     string
+
+	Message string
+}
+
+// NewResult constructs new Result based on raw value.
+func NewResult(value map[string]interface{}) Result {
+	return Result{
```
chen-keinan, line 64: same here value["id"] should be checked for existance

pkg/configauditreport/controller.go
```diff
@@ -0,0 +1,437 @@
...
+		return v1alpha1.ConfigAuditReportData{}, err
+	}
+
+	summary := v1alpha1.ConfigAuditSummary{}
+
+	for _, check := range checks {
+		if check.Success {
+			continue
+		}
+		switch check.Severity {
```
chen-keinan, line 283: I would export the severity switch  check to a util , is it commonly used , we have a similar code in :
https://github.com/aquasecurity/starboard/blob/main/pkg/plugin/polaris/plugin.go#L522

pkg/operator/etc/config.go
```diff
@@ -25,15 +25,31 @@ type Config struct {
...
+	LeaderElectionID      string `env:"OPERATOR_LEADER_ELECTION_ID" envDefault:"starboard-lock"`
 }
 
 // GetOperatorConfig loads Config from environment variables.
 func GetOperatorConfig() (Config, error) {
 	var config Config
 	err := env.Parse(&config)
+
+	if config.ConfigAuditScannerEnabled && config.ConfigAuditScannerBuiltIn {
+		return Config{}, fmt.Errorf("you cannot enabled both plugin-based and built-in configuration audit scanner")
```
chen-keinan, line 50: maybe replace with objective message  `plugin-based and built-in configuration audit scanners cannot be enabled at the same time`

pkg/policy/policy.go
danielpacak: Good point. I added checks for nil and blank strings for mandatory properties.

pkg/policy/policy.go
danielpacak: Done

pkg/configauditreport/controller.go
danielpacak: I've extracted this logic to the `v1alpha1.ConfigAuditSummaryFromChecks(checks []v1alpha1.Check)` method.

pkg/operator/etc/config.go
danielpacak: Done

chen-keinan (approved):

PR #985, opened 2022-02-22
feat(starboard-scanner-aqua): Add support for fs scan option

deven0t:
1. starboard-scanner-aqua shim to support filesystem command, so that scannercli fs-scan commnad is executed.
2. refactor code to get auth-key from aqua console api url

### DIFF
--- a/pkg/plugin/aqua/client/client.go
+++ b/pkg/plugin/aqua/client/client.go
@@ -1,6 +1,7 @@
 package client
 
 import (
+	"encoding/base64"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -18,6 +19,7 @@ var ErrUnauthorized = errors.New("unauthorized")
 
 type client struct {
 	baseURL       string
+	authHeader    string
 	authorization Authorization
 	httpClient    *http.Client
 }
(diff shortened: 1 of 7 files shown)


pkg/plugin/aqua/client/client.go
```diff
@@ -30,11 +32,36 @@ func (c *client) newGetRequest(url string) (*http.Request, error) {
 	req.Header.Add("Content-Type", "application/json; charset=UTF-8")
 	req.Header.Add("User-Agent", userAgent)
 	if auth := c.authorization.Basic; auth != nil {
-		req.SetBasicAuth(auth.Username, auth.Password)
+		req.Header.Add(c.authHeader, "Basic "+base64.StdEncoding.EncodeToString([]byte(auth.Username+":"+auth.Password)))
```
chen-keinan, line 35: maybe use formatting instead of concatenating

pkg/plugin/aqua/client/client.go
```diff
@@ -30,11 +32,36 @@ func (c *client) newGetRequest(url string) (*http.Request, error) {
...
+	url := fmt.Sprintf("%s/api", c.baseURL)
+	req, err := http.NewRequest(http.MethodGet, url, nil)
+	if err != nil {
+		return
+	}
+	req.Header.Add("Content-Type", "application/json; charset=UTF-8")
+	req.Header.Add("User-Agent", userAgent)
+	resp, err := c.httpClient.Do(req)
+	if err != nil {
+		return
```
chen-keinan, line 53: error is not handle or return

pkg/plugin/aqua/client/client.go
```diff
@@ -68,6 +95,7 @@ func NewClient(baseURL string, authorization Authorization) *Client {
 		authorization: authorization,
 		httpClient:    httpClient,
 	}
+	client.setAuthHeader()
```
chen-keinan, line 98: maybe its better to take out auth setting from client constructor , so error can be handled

danielpacak: It would be easier to read this code if we renamed setAuthHeader to getAuthHeader and let it return an error for proper error handling. Then we can implement the fallback logic, e.g. default to Authorization header. It's also fine to let NewClient return the error.

BTW, is it going to work if we fallback to Authorization header?

danielpacak (changes requested):
This is looking really nice @deven0t See our comments to handle errors when we resolve authentication header in the API client.

pkg/plugin/aqua/client/client.go
deven0t: thanks Chen, I have updated this.

PR #986, opened 2022-02-25
feat(aqua_scanner): aqua scanner plugin with filesystem scan option

deven0t:
Added support for aqua.command setting to use filesystem(fs-scan) scan option
from aqua enterprise scanner.


pkg/plugin/aqua/plugin.go
```diff
@@ -129,59 +261,156 @@ func (s *plugin) newScanJobContainer(ctx starboard.PluginContext, podContainer c
...
+	aquaImageRef, err := s.getImageRef(ctx)
+	if err != nil {
+		return corev1.PodSpec{}, nil, err
+	}
+	starboardAquaImage, err := config.GetStarboardAquaScannerImage()
+	if err != nil {
+		return corev1.PodSpec{}, nil, err
+	}
+	secretName := vulnerabilityreport.GetScanJobName(object) + "-volume"
+	var env []corev1.EnvVar
```
chen-keinan, line 304: the env param below is not used in `getPodSpecForFileSystemCommand` method 
```
var env []corev1.EnvVar
	envVars, err := s.getEnvFromConfig(ctx, secretName)
	if err != nil {
		return corev1.PodSpec{}, nil, err
	}
	env = append(env, envVars...)```

maybe it should be moved to newScanJobContainerFSCommand , it will reduce the amount of env var passed

deven0t: Thanks @chen-keinan , we will need to send secretName. but secretName will be less data as compared to slice of envVars. I will update PR.

deven0t: But I realized that we will need to ctx as well and env slice will be generated equal to the number of containers present in the original workload. So should we do that or leave it as current way.

chen-keinan: I see , then better keeping it the current way

danielpacak (approved):
LGTM!

PR #992, opened 2022-03-02
feat: update severity enum in config audit check

danielpacak:
The ConfigAuditReport used danger and warning enums to describe
severity of audit checks. These enum values where inherited from
Polaris data model. This patch updates severities to CRITICAL,
HIGH, MEDIUM, and LOW.


pkg/plugin/conftest/plugin.go
```diff
@@ -393,7 +393,7 @@ func (p *plugin) ParseConfigAuditReportData(ctx starboard.PluginContext, logsRea
 		for _, warning := range cr.Warnings {
```
chen-keinan, line 393: probebly we need to remove this as well :
```
if cr.Successes > 0 {
			successesCount += cr.Successes
		}
```

danielpacak: Good catch. I have removed it.

chen-keinan (approved):

yurrriq:
Critical is spelled incorrectly here. See also #1117

PR #993, opened 2022-03-03
refactor: parse severity from name

danielpacak:

### DIFF
--- a/pkg/plugin/polaris/plugin.go
+++ b/pkg/plugin/polaris/plugin.go
@@ -434,11 +434,15 @@ func (p *plugin) ParseConfigAuditReportData(ctx starboard.PluginContext, logsRea
 		return v1alpha1.ConfigAuditReportData{}, fmt.Errorf("unexpected report results count, got: %d, want: %d", len(report.Results), 1)
 	}
 	for _, pr := range report.Results[0].PodResult.Results {
+		severity, err := v1alpha1.StringToSeverity(pr.Severity)
+		if err != nil {
+			return v1alpha1.ConfigAuditReportData{}, err
+		}
 		check := v1alpha1.Check{
 			ID:       pr.ID,
 			Message:  pr.Message,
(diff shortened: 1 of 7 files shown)


pkg/plugin/polaris/plugin.go
```diff
@@ -434,11 +434,15 @@ func (p *plugin) ParseConfigAuditReportData(ctx starboard.PluginContext, logsRea
 		return v1alpha1.ConfigAuditReportData{}, fmt.Errorf("unexpected report results count, got: %d, want: %d", len(report.Results), 1)
 	}
 	for _, pr := range report.Results[0].PodResult.Results {
+		severity, err := v1alpha1.StringToSeverity(pr.Severity)
+		if err != nil {
+			return v1alpha1.ConfigAuditReportData{}, err
```
chen-keinan, line 439: should we fail the report generation for unknown severity ? or maybe we should log the error and `continue`

pkg/plugin/polaris/plugin.go
```diff
@@ -448,11 +452,15 @@ func (p *plugin) ParseConfigAuditReportData(ctx starboard.PluginContext, logsRea
 	for _, cr := range report.Results[0].PodResult.ContainerResults {
 		var containerChecks []v1alpha1.Check
 		for _, crr := range cr.Results {
+			severity, err := v1alpha1.StringToSeverity(crr.Severity)
+			if err != nil {
+				return v1alpha1.ConfigAuditReportData{}, err
```
chen-keinan, line 457: same comment as above

pkg/plugin/polaris/plugin.go
danielpacak: If it's not danger or warning we cannot reliably map it to v1alpha1.Severity enum. This would indicate non-backward compatible change in Polaris output or other bug in our parser.

chen-keinan (approved):

PR #1004, opened 2022-03-07
feat: compliance report support

chen-keinan:
Signed-off-by: chen-keinan <hen.keinan@gmail.com>
design doc can be found here: #961


deploy/crd/clustercompliancereports.crd.yaml
```diff
@@ -0,0 +1,121 @@
...
+                      name:
+                        type: string
+                      description:
+                        type: string
+                      id: # control check id
+                        type: string
+                      resources:
+                        type: array
+                        items:
+                          type: string # example Node,Workload
```
danielpacak, line 89: In K8s API machinery terms, Node and Workload are usually called kinds.

deploy/crd/clustercompliancereports.crd.yaml
```diff
@@ -0,0 +1,121 @@
...
+                      - name
+                      - id
+                      - resources
+                      - mapping
+                    properties:
+                      name:
+                        type: string
+                      description:
+                        type: string
+                      id: # control check id
```
danielpacak, line 84: Instead of using comments we can explain the meaning of each property by describing it. See how it's done for vulnerabilityreports.
```suggestion
                      id:
                        description: "control check id"
```

This would allows us to explain the CRD with kubectl, e.g. describe the `cron` property:

```console
$ kubectl explain clustercompliancereports.spec.cron
KIND:     ClusterComplianceReport
VERSION:  aquasecurity.github.io/v1alpha1

FIELD:    cron <string>

DESCRIPTION:
     <empty>
```

deploy/crd/clustercompliancereports.crd.yaml
```diff
@@ -0,0 +1,121 @@
+apiVersion: apiextensions.k8s.io/v1
```
danielpacak, line 1: To simply the installation we have a single file deploy/static/starboard.yaml that should include these CRDs. You can update and use hack/update-starboard.yaml.sh script to concatenate installed resources.

deploy/helm/templates/specs/nsa-1.0.yaml
```diff
@@ -0,0 +1,266 @@
...
+  name: nsa
+spec:
+  specKind: compliance
+  name: nsa
+  description: National Security Agency - Kubernetes Hardening Guidance
+  version: "1.0"
+  cron: "* * * * *"
+  controls:
+    - name: Non-root containers
+      description: ''
```
danielpacak, line 14: If description is blank, can we simply delete it?

pkg/apis/aquasecurity/v1alpha1/compliance_types.go
```diff
@@ -0,0 +1,76 @@
...
+}
+
+//SpecCheck represent the tool who perform the control check
+type SpecCheck struct {
+	ID string `json:"id"`
+}
+
+//Mapping represent the tool who perform the control check
+type Mapping struct {
+	Tool   string      `json:"tool"`
```
danielpacak, line 49: We usually call tools used by Starboard as scanners. See v1alpha1.Scanner struct.

pkg/operator/etc/config.go
```diff
@@ -23,6 +23,7 @@ type Config struct {
 	HealthProbeBindAddress                       string         `env:"OPERATOR_HEALTH_PROBE_BIND_ADDRESS" envDefault:":9090"`
 	CISKubernetesBenchmarkEnabled                bool           `env:"OPERATOR_CIS_KUBERNETES_BENCHMARK_ENABLED" envDefault:"true"`
 	VulnerabilityScannerEnabled                  bool           `env:"OPERATOR_VULNERABILITY_SCANNER_ENABLED" envDefault:"true"`
+	ClusterComplianceEnabled                     bool           `env:"OPERATOR_CLUSTER_COMPLIANCE_ENABLED" envDefault:"true"`
```
danielpacak, line 26: Can we keep config params related to vulnerability scanning together? This new param is added in between.

pkg/plugin/conftest/plugin_test.go
```diff
@@ -1,4 +1,4 @@
-package conftest_test
+package conftest
```
danielpacak, line 1: Why you change the package name? It's a good practice to use `conftest_test` to make sure we are testing public APIs exposed by the `conftest` package, not private methods or implementation details.

pkg/plugin/conftest/plugin_test.go
```diff
@@ -819,6 +818,35 @@ func TestPlugin_ConfigHash(t *testing.T) {
...
+		equal  bool
+		want   string
+	}{
+		{name: "get Policy ID From Result", result: Result{Message: "aaa", Metadata: map[string]interface{}{"id": "1234"}}, want: "1234", equal: true},
+		{name: "get Policy Title From Result", result: Result{Message: "bbb", Metadata: map[string]interface{}{"title": "title 1234"}}, want: "title 1234", equal: true},
+		{name: "get Policy ig generator From Result", result: Result{Message: "ccc", Metadata: map[string]interface{}{}}, want: "", equal: false}}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got := a.getPolicyTitleFromResult(tt.result)
```
danielpacak, line 839: Testing private methods is usually where bigger problems with testability and well defined APIs start. We should validate the logic implemented in the getPolicyTitleFromResult method by exercising public APIs.

deploy/helm/templates/rbac.yaml
```diff
@@ -141,13 +141,21 @@ rules:
       - configauditreports
       - clusterconfigauditreports
       - ciskubebenchreports
+      - clustercompliancereports
```
danielpacak, line 144: We need to update deploy/static/02-starboard-operator.rbac.yaml and deploy/static/starboard.yaml as well. See also my comment regarding the hack/update-starboard.yaml.sh script.

danielpacak (changes requested):
Good progress @chen-keinan I had flick through the code and wanted to install it in my cluster, but I bumped into some issues with static deployment descriptors (i.e. compliance CRDs are not installed). Most importantly, let's make sure deploy/static/starboard.yaml is up to date before I continue reviewing the code.

pkg/plugin/conftest/plugin_test.go
```diff
@@ -734,6 +669,148 @@ func TestPlugin_ParseConfigAuditReportData(t *testing.T) {
 	}))
 }
 
+func TestPlugin_ParseConfigAuditReportDataWithTitleAndIDOnMetadata(t *testing.T) {
```
danielpacak, line 672: In Go, I'd typically group test cases as subtests or use table driven test cases:

```go
func ParseConfigAuditReportData(t *testing.T) {

	t.Run("WithTitleOnMetadataOnly", func(t *testing.T) {
		// ...
	});

	t.Run("WithTitleAndIDOnMetadata", func(t *testing.T) {
		// ...
	});
}
```

pkg/utils/dateutil.go
```diff
@@ -0,0 +1,33 @@
...
+}
+
+//DurationExceeded  check if duration is now meaning zero
+func DurationExceeded(duration time.Duration) bool {
+	return duration.Nanoseconds() <= 0
+}
+
+//timeToExpiration  return the duration between time to expiration
+func timeToExpiration(expiresAt time.Time) time.Duration {
+	return expiresAt.Sub(ext.NewSystemClock().Now())
```
danielpacak, line 27: You cannot reliably unit tests this code if you use clock system implementation directly. We should be using ext.Clock interface if we want to unit test date/time operations.

pkg/operator/controller/ttl_report.go
```diff
@@ -63,12 +64,8 @@ func (r *TTLReportReconciler) reconcileReport() reconcile.Func {
 		if err != nil {
 			return ctrl.Result{}, fmt.Errorf("failed parsing %v with value %v %w", v1alpha1.TTLReportAnnotation, ttlReportAnnotationStr, err)
 		}
-		creationTime := report.Report.UpdateTimestamp
-		ttlExpired, durationToTTLExpiration, err := ttlIsExpired(reportTTLTime, creationTime.Time)
-		if err != nil {
-			return ctrl.Result{}, err
-		}
-		if ttlExpired {
+		durationToTTLExpiration := ttlIsExpired(reportTTLTime, report.Report.UpdateTimestamp.Time)
```
danielpacak, line 67: Do we need these changes in scope of this PR? Maybe we can do refactor PR to limit the "blast radius". There are many changes already to review.

danielpacak (changes requested):
I do have couple of comments and suggestions to be resolved before I jump on reviewing the actual logic and tests.

pkg/operator/controller/ttl_report.go
chen-keinan: made it to avoid duplicate code

deploy/helm/templates/specs/nsa-1.0.yaml
```diff
@@ -0,0 +1,258 @@
...
+          - id: 1.2.23
+    - name: service mesh is configure
+      description: 'control check whether service mesh is used in cluster'
+      id: '9.0'
+      resources:
+        - Node
+      mapping:
+        scanner: kube-bench
+        checks:
+          - id: "<check need to be added>"
```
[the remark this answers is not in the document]
chen-keinan: yes

pkg/utils/dateutil.go
danielpacak: But how you inject Fixed time implementation for testing? It's like using time.Now() directly.

pkg/compliance/io.go
```diff
@@ -0,0 +1,249 @@
...
+		log:    log,
+	}
+}
+
+type cm struct {
+	client client.Client
+	log    logr.Logger
+}
+
+type summaryTotal struct {
```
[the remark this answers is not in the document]
chen-keinan: its used for populating data to compliance and complianceDetails report prefer not to couple it to compliance summary also its a mid logic calc I think it should populate data to summary at the end before publishing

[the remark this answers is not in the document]
danielpacak: Interesting, but both structs have the same fields so I don't see any point in defining. Would you also define `internal_string` instead of using built-in `string` in Go?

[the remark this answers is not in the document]
chen-keinan: true it has the same fields now but we want to keep it flexible, if I want to change it , since it is used for mid processing I do not need to bread the CRD structure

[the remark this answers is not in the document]
danielpacak: What flexibility we need to count totals? If v1alpha1.ComplianceSummary won't be enough then we can add custom struct.

pkg/apis/aquasecurity/v1alpha1/compliance_types.go
```diff
@@ -0,0 +1,76 @@
...
+}
+
+// ControlCheck provides the result of conducting a single audit step.
+type ControlCheck struct {
+	ID          string `json:"id"`
+	Name        string `json:"name"`
+	Description string `json:"description,omitempty"`
+	PassTotal   int    `json:"passTotal"`
+	FailTotal   int    `json:"failTotal"`
+	Severity    string `json:"severity"`
```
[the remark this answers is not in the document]
chen-keinan: the severity will be defined as part of control spec in a separate PR

deploy/crd/clustercompliancereports.crd.yaml
```diff
@@ -0,0 +1,127 @@
...
+                - specKind
+                - name
+                - description
+                - version
+                - cron
+                - controls
+              properties:
+                specKind:
+                  type: string
+                  description: 'spedKind define the spec type , example: compliance'
```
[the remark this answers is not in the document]
chen-keinan: will fix it in a separate  PR

deploy/crd/clustercompliancereports.crd.yaml
```diff
@@ -0,0 +1,127 @@
...
+                name:
```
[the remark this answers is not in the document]
chen-keinan: will fix it in a separate  PR

pkg/apis/aquasecurity/v1alpha1/common_types.go
```diff
@@ -55,3 +56,16 @@ type Scanner struct {
 	// Version the version of the scanner.
 	Version string `json:"version"`
 }
+
+// Compliance is the specs for a security assessment report.
+type Compliance struct {
+	// Name the name of the compliance report.
+	Kind string `json:"specKind"`
+	// Name the name of the compliance report.
+	Name string `json:"name"`
```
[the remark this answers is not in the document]
chen-keinan: will fix it in a separate PR

danielpacak (approved):

PR #1015, opened 2022-03-10
fix: use ttl check with date util

chen-keinan:
- consolidate ttl checks with date util to avoid code duplication


pkg/operator/controller/ttl_report.go
```diff
@@ -63,12 +66,8 @@ func (r *TTLReportReconciler) reconcileReport() reconcile.Func {
 		if err != nil {
 			return ctrl.Result{}, fmt.Errorf("failed parsing %v with value %v %w", v1alpha1.TTLReportAnnotation, ttlReportAnnotationStr, err)
 		}
-		creationTime := report.Report.UpdateTimestamp
-		ttlExpired, durationToTTLExpiration, err := ttlIsExpired(reportTTLTime, creationTime.Time)
-		if err != nil {
-			return ctrl.Result{}, err
-		}
-		if ttlExpired {
+		durationToTTLExpiration := utils.NextIntervalDuration(reportTTLTime, report.Report.UpdateTimestamp.Time, r.Clock)
```
danielpacak, line 69: TBH, this code was much more readable with `ttlIsEExpired` method that accepted 2 args. Do you think we can add `utils.IsTTLExpired(clock ext.Clock, creationTime time.Time, ttl time.Duration) (bool, time.Duration)` to the utils package?

pkg/utils/dateutil.go
```diff
@@ -26,3 +26,8 @@ func DurationExceeded(duration time.Duration) bool {
 func timeToExpiration(expiresAt time.Time, clock ext.Clock) time.Duration {
 	return expiresAt.Sub(clock.Now())
 }
+
+//NextIntervalDuration  check if interval for given time has exceeded
+func NextIntervalDuration(interval time.Duration, creationTime time.Time, clock ext.Clock) time.Duration {
```
danielpacak, line 31: I found this method name and GDoc quite misleading.

danielpacak (changes requested):

pkg/operator/controller/ttl_report.go
chen-keinan: sure, I prefer to add it to ttl_report controller as dateUtil should be agnostic to logic terms (a.k.a: ttl)

chen-keinan: sure can do that, initially didn't want date util to include logic terms (ttl)

danielpacak (approved):

PR #1016, opened 2022-03-10
fix: remove compliance specKind field

chen-keinan:
remove  specKind field from compliance reports

### DIFF
--- a/pkg/compliance/io.go
+++ b/pkg/compliance/io.go
@@ -105,7 +105,10 @@ func (w *cm) createComplianceDetailReport(ctx context.Context, spec v1alpha1.Rep
 		ObjectMeta: metav1.ObjectMeta{
 			Name: name,
 		},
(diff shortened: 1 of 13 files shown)


pkg/compliance/io.go
```diff
@@ -105,7 +105,10 @@ func (w *cm) createComplianceDetailReport(ctx context.Context, spec v1alpha1.Rep
 		ObjectMeta: metav1.ObjectMeta{
 			Name: name,
 		},
-		Report: v1alpha1.ClusterComplianceDetailReportData{UpdateTimestamp: metav1.NewTime(ext.NewSystemClock().Now()), Summary: summary, Type: v1alpha1.Compliance{Kind: strings.ToLower(spec.Kind), Name: name, Description: strings.ToLower(spec.Description), Version: spec.Version}, ControlChecks: controlChecksDetails},
+		Report: v1alpha1.ClusterComplianceDetailReportData{UpdateTimestamp: metav1.NewTime(ext.NewSystemClock().Now()),
```
danielpacak, line 108: It's not directly related to this PR, but I spot it right now.

If you use `ext.NewSystemClock().Now()` instead of injecting `ext.Clock` interface doesn't make much sense. You could just use `time.Now()` and keep this code untestable.

The `ext.Clock` interface was introduce to easily swap implementation of the clock in unit tests from `ext.NewSystemClock()` to `ext.NewFixedClock(fixedTime time.Time)` with dependency injection pattern.

danielpacak (approved):

PR #1030, opened 2022-03-14
Feat/add compliance check default status

### LINKED ISSUE #1023: Add compliance check default status
**The requirement:** 
Add compliance check default status.

**Details:**
As part of the compliance check starboard, it requires validating the existence of a resource.
The Operator is an event-based system, therefore we can't query for resource existence we only react to events.

Adding a default status to compliance checks (optional) will help to sort out the check for the resource existence, meaning if starboard didn't get an event for a specific resource and the default status is marked as `fail`, then the compliance check will be marked as `fail` as a resource does not exist.

If starboard got an event for resources, the default status can be ignored and the compliance check result will be determined base on the resource scanning result
Fixed by: #1030

chen-keinan:
Add  support to compliance spec default status:
fixes #1023

### DIFF
--- a/deploy/crd/clustercompliancereports.crd.yaml
+++ b/deploy/crd/clustercompliancereports.crd.yaml
@@ -114,12 +114,13 @@ spec:
                           - MEDIUM
                           - LOW
                           - UNKNOWN
-                      defaultValue:
+                      defaultStatus:
                         type: string
                         description: 'define the default value for check status in case resource not found'
                         enum:
-                          - FAIL
                           - PASS
+                          - WARN
+                          - FAIL
             status:
               x-kubernetes-preserve-unknown-fields: true
               type: object
--- a/pkg/apis/aquasecurity/v1alpha1/compliance_types.go
+++ b/pkg/apis/aquasecurity/v1alpha1/compliance_types.go
@@ -31,13 +31,13 @@ type ReportSpec struct {
 
 //Control represent the cps controls data and mapping checks
 type Control struct {
-	ID           string       `json:"id"`
-	Name         string       `json:"name"`
-	Description  string       `json:"description,omitempty"`
-	Kinds        []string     `json:"kinds"`
-	Mapping      Mapping      `json:"mapping"`
-	Severity     Severity     `json:"severity"`
-	DefaultValue DefaultValue `json:"defaultValue,omitempty"`
+	ID            string        `json:"id"`
+	Name          string        `json:"name"`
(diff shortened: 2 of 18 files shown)


deploy/crd/clustercompliancereports.crd.yaml
```diff
@@ -114,6 +114,12 @@ spec:
                           - MEDIUM
                           - LOW
                           - UNKNOWN
+                      defaultStatus:
+                        type: string
+                        description: 'define the default value for check status in case resource not found'
+                        enum:
+                          - FAIL
```
danielpacak, line 121: Do we plan to add more enum values here? If not I'm wondering why don't we use boolean type?

pkg/apis/aquasecurity/v1alpha1/compliance_types.go
```diff
@@ -74,3 +75,10 @@ type ControlCheck struct {
 	FailTotal   int      `json:"failTotal"`
 	Severity    Severity `json:"severity"`
 }
+
+type DefaultStatus string
```
danielpacak, line 79: See my comment about using boolean instead of enum. If you want to proceed with enum then we can may call it ControlStatus? And then use it as type of the DefaultStatus property.

deploy/crd/clustercompliancereports.crd.yaml
chen-keinan: I prefer use enum as it can be 'FAIL' / 'PASS' or not define , if we use boolean default value is false

danielpacak: Interesting. Cannot we override the default boolean value from false to true? Currently you always set the default value with string literals.

chen-keinan: I mean, if you do not want to have default value for specific control check ?

danielpacak: I'm asking about enum and whether it's absolutely necessary to define enum that looks like built-in boolean data type.

danielpacak: See how it's done for some built-in Kubernetes resources such as ServiceAccounts or PodSpec definition:
```go
import (
	"k8s.io/utils/pointer"
)

serviceAccount := &corev1.ServiceAccount{
		Name: starboard.ServiceAccountName,
		AutomountServiceAccountToken: pointer.BoolPtr(false),
	},
}
```

danielpacak (approved):

PR #1033, opened 2022-03-16
docs: add nsa 1.0 spec

chen-keinan:
Add NSA Kubernetes Hardening Guidance 1.0


danielpacak (commented):
Why do we add it here instead of linking to NSA website? Do they host only the latest version of spec?

chen-keinan:
@danielpacak yes , its look like they keep only the latest spec

danielpacak (approved):

PR #1037, opened 2022-03-17
feat: using oci image spec annotations

krol3:
The official Starboard images are labeled following [the org.label-schema Label Schema](https://label-schema.org/). That schema has been deprecated in favor of [the superseeding OCI image spec annotations](https://github.com/opencontainers/image-spec/blob/main/annotations.md#back-compatibility-with-label-schema).


danielpacak (approved):
Thank you @krol3 !

PR #1066, opened 2022-03-24
Feat/automate install resources in CLI

### LINKED ISSUE #1065: automate resource installation within CLI
The motivation for this change is to simplify CLI experience amount of operations needed to be done to get scan results:
- autodetect if resources have been installed when running the `get`, `scan`, and `reports` commands
   - if resources are not installed then install them and proceed with the command
   -  if resources are installed then proceed with the command
- need to make sure the resources are updated upon CLI upgrade 
- keep the install & uninstall command to have a proper cleanup and computability

Fixed by: #1066 

***Details:*** 

Before this change, we needed to run the `starboard install` `starboard install`  command before we used the `scan` or `get` commands. 

After this change, we can use right away the `scan` and `get` commands without the need to run the `install` command, behind the scenes, the installation will happen as a Pre-command action for scan / get

***Note:*** the `starboard uninstall` command will remain, as we still need a way to clean up starboard resources

For now, we will keep the `starboard install` command, (can be deprecated later if needed) but it will be removed from documentation to simplify the experience of CLI

chen-keinan:
Fixes #1065

### DIFF
--- a/pkg/cmd/metadata.go
+++ b/pkg/cmd/metadata.go
@@ -1,0 +1,167 @@
+package cmd
+
+import (
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"os/user"
+	"path"
+	"path/filepath"
+	"time"
+
+	"github.com/aquasecurity/starboard/pkg/starboard"
+	"github.com/spf13/cobra"
+	"k8s.io/cli-runtime/pkg/genericclioptions"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	starboardHomeFolder = ".starboard"
+	MetadataFileName    = "metadata.json"
+)
+
+func createOrUpdateResourcesAndMetadata(buildInfo starboard.BuildInfo, cf *genericclioptions.ConfigFlags) func(cmd *cobra.Command, args []string) error {
+	return func(cmd *cobra.Command, args []string) error {
+		filePath, err := getMetadataFilePath()
+		if err != nil {
+			return err
+		}
+		if _, err := os.Stat(filePath); os.IsNotExist(err) {
+			err := createResourcesAndMetadata(buildInfo, cf, filePath)
+			if err != nil {
+				return err
+			}
+		} else {
+			err := updateResourcesAndMetadata(filePath, buildInfo, cf)
+			if err != nil {
+				return err
+			}
+		}
+		return nil
+	}
+}
+
+func updateResourcesAndMetadata(filePath string, buildInfo starboard.BuildInfo, cf *genericclioptions.ConfigFlags) error {
+	if needUpdate, err := needUpdate(filePath, buildInfo); err == nil && needUpdate {
+		err := deleteStarboardMetadataFile(filePath)
+		if err != nil {
+			return err
+		}
+		err = createResourcesAndMetadata(buildInfo, cf, filePath)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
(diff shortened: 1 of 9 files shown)


pkg/cmd/metadata.go
```diff
@@ -0,0 +1,167 @@
...
+	}
+}
+
+func updateResourcesAndMetadata(filePath string, buildInfo starboard.BuildInfo, cf *genericclioptions.ConfigFlags) error {
+	if needUpdate, err := needUpdate(filePath, buildInfo); err == nil && needUpdate {
+		err := deleteStarboardMetadataFile(filePath)
+		if err != nil {
+			return err
+		}
+		err = createResourcesAndMetadata(buildInfo, cf, filePath)
```
josedonizetti, line 51:
```suggestion
		return createResourcesAndMetadata(buildInfo, cf, filePath)
```

pkg/cmd/metadata.go
```diff
@@ -0,0 +1,167 @@
...
+	scheme := starboard.NewScheme()
+	kubeClient, err := client.New(kubeConfig, client.Options{Scheme: scheme})
+	if err != nil {
+		return err
+	}
+	err = installResources(kubeClient, kubeConfig, buildInfo)
+	if err != nil {
+		return err
+	}
+	err = createMetadataFile(buildInfo, err, filePath)
```
josedonizetti, line 74:
```suggestion
	return createMetadataFile(buildInfo, err, filePath)
```

pkg/cmd/metadata.go
```diff
@@ -0,0 +1,167 @@
...
+	return filepath.Join(folderPath, MetadataFileName), nil
+}
+
+func deleteStarboardMetadataFolder() func(cmd *cobra.Command, args []string) error {
+	return func(cmd *cobra.Command, args []string) error {
+		starboardHomeFolder, err := getStarboardHomeFolder()
+		if err != nil {
+			return err
+		}
+		err = os.RemoveAll(starboardHomeFolder)
```
josedonizetti, line 128:
```suggestion
		remove os.RemoveAll(starboardHomeFolder)
```

pkg/cmd/metadata.go
```diff
@@ -0,0 +1,167 @@
...
+		if err != nil {
+			return err
+		}
+		return nil
+	}
+}
+
+func deleteStarboardMetadataFile(starboardMetadataFile string) error {
+	err := os.Remove(starboardMetadataFile)
```
josedonizetti, line 137:
```suggestion
	remove os.Remove(starboardMetadataFile)
```

pkg/cmd/metadata.go
```diff
@@ -0,0 +1,167 @@
...
+type Metadata struct {
+	Version     string `json:"version"`
+	Initialized bool   `json:"initialized"`
+	CreatedAt   string `json:"createdAt"`
+}
+
+//createStarboardHomeFolderIfNotExist create starboard home folder if not exist
+func createStarboardHomeFolderIfNotExist(folderName string) error {
+	if _, err := os.Stat(folderName); os.IsNotExist(err) {
+		errDir := os.MkdirAll(folderName, 0700)
```
josedonizetti, line 153:
```suggestion
		return os.MkdirAll(folderName, 0700)
```

josedonizetti (commented):
@chen-keinan I did some simple suggestions to remove some `if err`, nothing big, just style feel free to ignore. One thing I'm not sure if we have other issue to tackle it already or not, is the feedback from the install precommand:
```
bin/starboard scan vulnerabilityreports deployment/nginx
updating starboard resources...
```
Saying only `updating starboard resources...` is too vague, I would prefer knowing exactly what was installed on the cluster. WDYT?

PR #1108, opened 2022-04-04
fix: limit the num of failure entries in compliance detail report

### LINKED ISSUE #1106: Limit the num of failure entries per control check in cluster compliance detail report
***The problem:***
Limiting the number of failure entries per control check on the cluster compliance detail report is required.
Especially in mid/big clusters when there are many resources with failures will produce a huge cluster compliance detail report that exceeds the etcd limits.

***Suggested resolution:***
it is suggested to limit, by configuration, the number of failures entries per control check in the cluster compliance detail report 

Fixed by: #1108

chen-keinan:
Signed-off-by: chenk <hen.keinan@gmail.com>

## Description
This change limits the number of fail entries per control check in the cluster compliance detail report.

## Related issues
- Close #1106 

## Checklist
- [x] I've read the [guidelines for contributing](https://github.com/aquasecurity/starboard/blob/main/CONTRIBUTING.md) to this repository.
- [x] I've added tests that prove my fix is effective or that my feature works.
- [x] I've updated the [documentation](https://github.com/aquasecurity/starboard/tree/main/docs) with the relevant information (if needed).
- [x] I've added usage information (if the PR introduces new options)

### DIFF
--- a/deploy/helm/values.yaml
+++ b/deploy/helm/values.yaml
@@ -47,8 +47,10 @@ operator:
   configAuditScannerBuiltIn: true
   # kubernetesBenchmarkEnabled the flag to enable CIS Kubernetes Benchmark scanner
   kubernetesBenchmarkEnabled: true
-  # clusterComplianceEnabled the flag to enable CIS Kubernetes Benchmark scanner
+  # clusterComplianceEnabled the flag to enable cluster compliance report generation
   clusterComplianceEnabled: true
+  # ClusterComplianceFailEntriesLimit the flag to limit the number of fail entries per control check in the cluster compliance detail report
+  ClusterComplianceFailEntriesLimit: 10
   # batchDeleteLimit the maximum number of config audit reports deleted by the operator when the plugin's config has changed.
   batchDeleteLimit: 10
   # vulnerabilityScannerScanOnlyCurrentRevisions the flag to only create vulnerability scans on the current revision of a deployment.
--- a/deploy/helm/values.yaml
+++ b/deploy/helm/values.yaml
@@ -49,8 +49,6 @@ operator:
   kubernetesBenchmarkEnabled: true
   # clusterComplianceEnabled the flag to enable cluster compliance report generation
   clusterComplianceEnabled: true
-  # ClusterComplianceFailEntriesLimit the flag to limit the number of fail entries per control check in the cluster compliance detail report
-  ClusterComplianceFailEntriesLimit: 10
   # batchDeleteLimit the maximum number of config audit reports deleted by the operator when the plugin's config has changed.
   batchDeleteLimit: 10
   # vulnerabilityScannerScanOnlyCurrentRevisions the flag to only create vulnerability scans on the current revision of a deployment.
@@ -178,7 +176,9 @@ trivy:
   # Trivy client to Trivy server. Only applicable in ClientServer mode.
   #
   # serverCustomHeaders: "foo=bar"
-
+compliance:
+  # failEntriesLimit the flag to limit the number of fail entries per control check in the cluster compliance detail report
+  failEntriesLimit: 10
 kubeBench:
   imageRef: docker.io/aquasec/kube-bench:v0.6.6
 
--- a/docs/operator/configuration.md
+++ b/docs/operator/configuration.md
@@ -22,7 +22,8 @@ Configuration of the operator's Pod is done via environment variables at startup
 | `OPERATOR_VULNERABILITY_SCANNER_REPORT_TTL`                  | `""`                 | The flag to set how long a vulnerability report should exist. When a old report is deleted a new one will be created by the controller. It can be set to `""` to disabled the TTL for vulnerability scanner. |
 | `OPERATOR_LEADER_ELECTION_ENABLED`                           | `false`              | The flag to enable operator replica leader election                                                                                                                                                          |
 | `OPERATOR_LEADER_ELECTION_ID`                                | `starboard-lock`     | The name of the resource lock for leader election                                                                                                                                                            |
-| `OPERATOR_CLUSTER_COMPLIANCE_ENABLED `                       | `true`               | The flag to enable Cluster Compliance report generation                                                                                                                                                      |
+| `OPERATOR_CLUSTER_COMPLIANCE_ENABLED`                        | `true`               | The flag to enable Cluster Compliance report generation                                                                                                                                                      |
+| `OPERATOR_CLUSTER_COMPLIANCE_FAIL_ENTRIES_LIMIT`             | `10`                 | The flag to limit the number of fail entries per control check in the cluster compliance detail report                                                                                                       |
 
 ## Install Modes
 
--- a/docs/operator/configuration.md
+++ b/docs/operator/configuration.md
@@ -22,8 +22,7 @@ Configuration of the operator's Pod is done via environment variables at startup
 | `OPERATOR_VULNERABILITY_SCANNER_REPORT_TTL`                  | `""`                 | The flag to set how long a vulnerability report should exist. When a old report is deleted a new one will be created by the controller. It can be set to `""` to disabled the TTL for vulnerability scanner. |
 | `OPERATOR_LEADER_ELECTION_ENABLED`                           | `false`              | The flag to enable operator replica leader election                                                                                                                                                          |
 | `OPERATOR_LEADER_ELECTION_ID`                                | `starboard-lock`     | The name of the resource lock for leader election                                                                                                                                                            |
-| `OPERATOR_CLUSTER_COMPLIANCE_ENABLED`                        | `true`               | The flag to enable Cluster Compliance report generation                                                                                                                                                      |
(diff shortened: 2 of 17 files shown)


deploy/helm/values.yaml
```diff
@@ -47,8 +47,10 @@ operator:
   configAuditScannerBuiltIn: true
   # kubernetesBenchmarkEnabled the flag to enable CIS Kubernetes Benchmark scanner
   kubernetesBenchmarkEnabled: true
-  # clusterComplianceEnabled the flag to enable CIS Kubernetes Benchmark scanner
+  # clusterComplianceEnabled the flag to enable cluster compliance report generation
   clusterComplianceEnabled: true
+  # ClusterComplianceFailEntriesLimit the flag to limit the number of fail entries per control check in the cluster compliance detail report
```
danielpacak, line 52:
```suggestion
  # ClusterComplianceFailEntriesLimit the number of fail entries per control check in the cluster compliance detail report
```

docs/operator/configuration.md
```diff
@@ -22,7 +22,8 @@ Configuration of the operator's Pod is done via environment variables at startup
...
 | `OPERATOR_LEADER_ELECTION_ID`                                | `starboard-lock`     | The name of the resource lock for leader election                                                                                                                                                            |
-| `OPERATOR_CLUSTER_COMPLIANCE_ENABLED `                       | `true`               | The flag to enable Cluster Compliance report generation                                                                                                                                                      |
+| `OPERATOR_CLUSTER_COMPLIANCE_ENABLED`                        | `true`               | The flag to enable Cluster Compliance report generation                                                                                                                                                      |
+| `OPERATOR_CLUSTER_COMPLIANCE_FAIL_ENTRIES_LIMIT`             | `10`                 | The flag to limit the number of fail entries per control check in the cluster compliance detail report                                                                                                       |
```
danielpacak, line 26: The flag is for booleans? Here we describe number.

pkg/cmd/get_clustercompliancereport.go
```diff
@@ -50,7 +52,11 @@ func NewGetClusterComplianceReportsCmd(executable string, cf *genericclioptions.
 				return err
 			}
 			// generate compliance and compliance failure detail reports
-			complianceMgr := compliance.NewMgr(kubeClient, logger)
+			operatorConfig, err := etc.GetOperatorConfig()
```
danielpacak, line 55: The CLI depends on the struct that's meant for the Operator and introduced the coupling between Go packages.

pkg/cmd/get_clustercompliancereport.go
```diff
@@ -50,7 +52,11 @@ func NewGetClusterComplianceReportsCmd(executable string, cf *genericclioptions.
...
+			if err != nil {
+				return fmt.Errorf("getting client config: %w", err)
```
danielpacak, line 57: The "client" is not very clear probably:
```suggestion
				return fmt.Errorf("getting operator config: %w", err)
```

pkg/compliance/clustercompliancereport_test.go
```diff
@@ -35,7 +35,8 @@ func loadResource(filePath string, resource interface{}) error {
 
 var _ = ginkgo.Describe("cluster compliance report", func() {
 	config := etc.Config{
```
danielpacak, line 37: I suggest that we don't require operator-specific types (especially whole config) in the compliance package.

pkg/compliance/io.go
```diff
@@ -5,6 +5,8 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/aquasecurity/starboard/pkg/operator/etc"
+
```
danielpacak, line 9:
```suggestion
```

docs/operator/configuration.md
danielpacak: Also if it's configurable for CLI and Operator, it should go into the starboard ConfigMap

pkg/compliance/io.go
```diff
@@ -23,16 +25,18 @@ type Mgr interface {
 	GenerateComplianceReport(ctx context.Context, spec v1alpha1.ReportSpec) error
 }
 
-func NewMgr(client client.Client, log logr.Logger) Mgr {
+func NewMgr(client client.Client, log logr.Logger, config etc.Config) Mgr {
```
danielpacak, line 28: Do you need to pass the whole operator config to the compliance manager? Maybe you can use [functional options](https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis) pattern here and just pass the limit config?
The compliance mgr should know nothing about the Operator and it's config IMO

danielpacak (changes requested):

deploy/helm/values.yaml
```diff
@@ -176,7 +176,9 @@ trivy:
   # Trivy client to Trivy server. Only applicable in ClientServer mode.
   #
   # serverCustomHeaders: "foo=bar"
-
+compliance:
+  # ClusterComplianceFailEntriesLimit the flag to limit the number of fail entries per control check in the cluster compliance detail report
```
danielpacak, line 180: I think this prefix is wrong here in the comment

pkg/cmd/get_clustercompliancereport.go
```diff
@@ -49,8 +51,15 @@ func NewGetClusterComplianceReportsCmd(executable string, cf *genericclioptions.
...
-			complianceMgr := compliance.NewMgr(kubeClient, logger)
+			kubeClientset, err := kubernetes.NewForConfig(kubeConfig)
+			if err != nil {
+				return err
+			}
+			starboardConfig, err := starboard.NewConfigManager(kubeClientset, starboard.NamespaceName).Read(ctx)
+			if err != nil {
+				return err
+			}
+			complianceMgr := compliance.NewMgr(kubeClient, logger, starboardConfig)
```
danielpacak, line 62: Accepting it for now, but as I mentioned in my first review, when you inject dependencies limit them to absolute minimum. Otherwise you are introducing unnecessary coupling between two Go packages, i.e. starboard and compliance. The starboard configuration object configured other things, not just compliance.

pkg/cmd/get_clustercompliancereport.go
```diff
@@ -3,11 +3,13 @@ package cmd
 import (
 	"context"
 	"fmt"
+	"io"
```
danielpacak, line 6: Use goimport here

pkg/compliance/clustercompliancereport_test.go
```diff
@@ -216,3 +216,28 @@ func getDetailReport(ctx context.Context, namespaceName types.NamespacedName, cl
 	}
 	return &report, nil
 }
+
+func getStarboardConfig() (starboard.ConfigData, error) {
+	clientset := fakesimple.NewSimpleClientset(
```
danielpacak, line 221: Why do you use fake here?? ConfigData is a Go struct that you can just instantiate easily.

```go
func getStarboardConfig() (starboard.ConfigData, error) {
	return starboard.ConfigData{
		"compliance.failEntriesLimit": "1",
	}, nil
}
```

You probably don't even need this private method and error in return param.

pkg/compliance/clustercompliancereport_test.go
```diff
@@ -216,3 +216,28 @@ func getDetailReport(ctx context.Context, namespaceName types.NamespacedName, cl
...
+				"compliance.failEntriesLimit": "1",
+			},
+		},
+		&corev1.Secret{
+			ObjectMeta: metav1.ObjectMeta{
+				Namespace: starboard.NamespaceName,
+				Name:      starboard.SecretName,
+			},
+			Data: map[string][]byte{
+				"baz": []byte("s3cret"),
```
danielpacak, line 237: Is this secret used somehow in the test scenarios? If it's not required there is no point in adding it to test fixture.

pkg/compliance/io.go
```diff
@@ -23,16 +25,18 @@ type Mgr interface {
 	GenerateComplianceReport(ctx context.Context, spec v1alpha1.ReportSpec) error
 }
 
-func NewMgr(client client.Client, log logr.Logger) Mgr {
+func NewMgr(client client.Client, log logr.Logger, config starboard.ConfigData) Mgr {
```
danielpacak, line 28: Like I said in my first review and I will repeat it here. Use functional options pattern or just config object specific to compliance checks. By injecting the whole starboard.ConfigData, which configures kube-bench, kube-hunter, default scanners, etc., you are introducing unnecessary coupling between Go packages.

danielpacak (changes requested):

pkg/compliance/io.go
chen-keinan: not sure I agree, the starboardConfig is injected to all (scanners ) plugins and controllers (vulnerability, cis-bench, and config audit). I do not see why this case is different

danielpacak: Because it's injected into existing code, it doesn't mean we must inject it into new code. Existing code is based on POC and experiments and our goal now is to write clean and maintainable code.

danielpacak (approved):

PR #1118, opened 2022-04-08
Changes for Starboard Operator with ubi8-minimal base Image Resolve: SLK-44766

prakhar-aqua:
This PR contains changes in Makefile and addition of Dockerfile with ubi8-minimal as base Image for OCP Certification of Starboard-Operator.


danielpacak:
Could you explain why would you add such change? Is it only for local development? If not, how would you expect us to release and publish to DockerHub for different architectures? We're using GoReleaser and these Makefile targets are only useful for development workflows.

PR #1144, opened 2022-04-20
feat(helm): support custom policies with built-in scanner

yurrriq:
## Description

Add support for custom policies when using the custom scanner.

## Related issues
- Close #1143

Remove this section if you don't have related PRs.

## Checklist
- [ ] I've read the [guidelines for contributing](https://github.com/aquasecurity/starboard/blob/main/CONTRIBUTING.md) to this repository.
- [ ] I've added tests that prove my fix is effective or that my feature works.
- [ ] I've updated the [documentation](https://github.com/aquasecurity/starboard/tree/main/docs) with the relevant information (if needed).
- [ ] I've added usage information (if the PR introduces new options)
- [ ] I've included a "before" and "after" example to the description (if the PR is a user interface change).


deploy/helm/values.yaml
```diff
@@ -468,3 +468,19 @@ nodeSelector: {}
 tolerations: []
 
 affinity: {}
+
+additionalPolicies:
```
yurrriq, line 472: As far as I know, the Conftest is deprecated and planned for removal. If that's _not_ the case, perhaps it would be good to unify this with `conftest.library` and `conftest.policy`.

yurrriq: I'm also not attached to this name, but it seems to follow in the tradition of other charts in the wild.

deploy/helm/templates/policies.yaml
```diff
@@ -863,3 +863,12 @@ data:
     kubernetes.namespace]))\n\n\tres := {\n\t\t\"msg\": msg,\n\t\t\"id\": __rego_metadata__.id,\n\t\t\"title\":
     __rego_metadata__.title,\n\t\t\"severity\": __rego_metadata__.severity,\n\t\t\"type\":
     __rego_metadata__.type,\n\t}\n}\n"
+  {{- with .Values.additionalPolicies }}
```
yurrriq, line 866: It would be interesting to try to prevent collisions here. Otherwise our beloved YAML will let the last instance of a duplicate key win, which might produce unexpected results.

yurrriq:
For my use case, this could be obviated by #1145, but I imagine others might prefer this simpler approach. Supporting both seems prudent.

PR #1269, opened 2022-09-19
Using latest tag everytime for ubi-minimal

kaljonwal:
Instead of using specific tag from ubi-minimal will use latest tag everytime. So all time will have all latest packages with fixed vulnerabilities.

Note: with ubi-minimal:8.5 we are seeing vulnerabilities.


chen-keinan:
LGTM 🚀

chen-keinan (approved):

PR #1352, opened 2023-05-25
feat: Support scan latest revision for replicationController from Dep…

### LINKED ISSUE #1351: OCP | Scan current revision is not supported for replicationController created from DeploymentConfig
Run Starboard operator with "OPERATOR_CONFIG_AUDIT_SCANNER_SCAN_ONLY_CURRENT_REVISIONS": true
So that it will scan only current revisions. 
This flag is getting applied only to replicaset from deployment and not to replicationController from deploymentConfig
Due to this configaudit report is getting generated for old revision as well, so we see many configauditreports for history revision.

[Miscellaneous information that will assist in solving the issue.]

**Environment:**
OCP

- Starboard version (use `starboard version`): 0.15.12
- Kubernetes version (use `kubectl version`): 1.22

deven0t:
…loymentConfig

In OCP, there is a kind called DeploymentConfig, which manages the replicationController, and as we have revisionHistory for Deployments, similarly there are revision history for Deployment Config as well. Due to which many replication contollers will be present in the cluster. We dont want to scan them as they are not active resource. So as part of this change, I have added following change
- consider "OPERATOR_CONFIG_AUDIT_SCANNER_SCAN_ONLY_CURRENT_REVISIONS" env var to limit scanning of old replication controller
- ingore pods created from deploymentConfig

Close #1351


chen-keinan:
lgtm 🚀

chen-keinan (approved):

Issue #1355, opened 2023-05-30
Error |  clusterRole with camelCase as name gives error in ClusterConfigAuditreport


deven0t:
- Deploy starboard with builtIn scanner for configaudit reports
- Create cluster role with camelCase or with some capital letters
- Then starboard fails to generate clusterconfigAuditReport and it throws following error 
`{"level":"error","ts":1685124736.0708067,"msg":"Reconciler error","controller":"clusterrolebinding","controllerGroup":"rbac.authorization.k8s.io","controllerKind":"ClusterRoleBinding","clusterRoleBinding":{"name":"system-netes-default-clusterRoleBinding"},"namespace":"","name":"system-netes-default-clusterRoleBinding","reconcileID":"ce611316-186c-4c4f-97f1-b12194ea6985","error":"ClusterConfigAuditReport.aquasecurity.github.io \"clusterrolebinding-system-netes-default-clusterRoleBinding\" is invalid: metadata.name: Invalid value: \"clusterrolebinding-system-netes-default-clusterRoleBinding\": a DNS-1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')","stacktrace":"sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem\n\t/home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.12.3/pkg/internal/controller/controller.go:273\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2\n\t/home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.12.3/pkg/internal/controller/controller.go:234"}  `

**Environment:**

- Starboard version (use `starboard version`): 0.15.10
- Kubernetes version (use `kubectl version`): 1.24

PR #1365, opened 2023-07-19
fix: Fix crd names error for roles

deven0t:
roles can have different chars which are not supported in crd names eg '_', capital letter

previous fix was only handling capital letter, this fix will handle other usecases as well.

fixes: #1355


chen-keinan:
lgtm

chen-keinan (approved):

PR #1367, opened 2023-08-18
opa version upgrade to latest

KiranBodipi:
## Description
To maintain compatibility with previous KE versions, we initially introduced the result. New feature to all Rego scripts on the server level. However, we encountered an error within StarBoard. This problem arises from the utilization of "dots in rule heads," which is not compatible with the current OPA version employed by StarBoard. As a resolution, updated the OPA version used within StarBoard to v0.46.1. Reference: https://github.com/open-policy-agent/opa/issues/4653
We were observing vulnerabilities in image, this PR will resolve the vulnerabilities as well.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -16,9 +16,9 @@ require (
 	github.com/hashicorp/go-version v1.5.0
 	github.com/onsi/ginkgo v1.16.5
 	github.com/onsi/gomega v1.20.0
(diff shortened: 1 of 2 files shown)


mjshastha (approved):
I'm content with the changes.

chen-keinan (approved):

chen-keinan (approved):

PR #1376, opened 2024-01-01
fix: Mark old revision configaudit report for deletion

deven0t:
Till now, when revision became old then old configaudit report was not deleted for it. So as part of this PR handled following changes
- Support report TTL for configaudit reports
- Mark old configaudit report for deletion
- Update controller-runtime pkg version


chen-keinan:
@deven0t lgtm 🚀

chen-keinan (approved):

PR #1405, opened 2024-11-06
perf: Optimize cache by minimizing unnecessary data storage.

mjshastha:
Updates to controller-runtime and client-go for better performance, bug fixes, and memory optimization.

Stripped managed fields and last-applied-configuration from resources and cache:
- Implemented object metadata transformation using `WithTransform` in `NewSharedInformerFactoryWithOptions` for shared informers. ref: informers package - k8s.io/client-go/informers - Go Packages
- Updated the controller runtime manager cache to include `DefaultTransform`, leveraging `TransformStripManagedFields()` along with logic to remove the last-applied-configuration annotation for all cached objects. ref: cache package - sigs.k8s.io/controller-runtime/pkg/cache - Go Packages
- These changes ensure that unwanted metadata is removed from objects before they are stored in the cache, optimizing memory usage and improving performance.

Additionally, fixed failing unit test cases.


itest/starboard-operator/behavior/behavior.go
```diff
@@ -36,34 +35,6 @@ type Inputs struct {
 func VulnerabilityScannerBehavior(inputs *Inputs) func() {
 	return func() {
 
-		Context("When unmanaged Pod is created", func() {
```
deven0t, line 39: any reason to remove this test case

itest/starboard-operator/behavior/behavior.go
```diff
@@ -79,15 +50,15 @@ func VulnerabilityScannerBehavior(inputs *Inputs) func() {
 
 				err := inputs.Create(ctx, deploy)
 				Expect(err).ToNot(HaveOccurred())
-				Eventually(inputs.HasActiveReplicaSet(inputs.PrimaryNamespace, deploy.Name), inputs.AssertTimeout).Should(BeTrue())
+				Eventually(inputs.HasActiveReplicaSet(inputs.PrimaryNamespace, deploy.Name), 300*time.Second).Should(BeTrue())
```
deven0t, line 53: I see we are updating AssertTimeout to 5min, so we dont need this change here. Let it come from inputs.Asserttimeout

mjshastha: will address in the next commit.

